PR #21, opened 2022-11-30
Feature/new article structure plugin

lagartoverde:
This PR converts the article-structure plugin with the new SHACL based API to prosemirror. It is more or less feature-par with the old version. In the future we should look into creating a new plugin with prosemirror in mind using specific prosemirror schemas and node-specs for the structures.

Should be tested in combination with https://github.com/lblod/ember-rdfa-editor/tree/nodes-between.

Solution to https://binnenland.atlassian.net/jira/software/c/projects/GN/boards/4?modal=detail&selectedIssue=GN-3865


addon/utils/article-structure-plugin/constants.ts
```diff
@@ -0,0 +1,260 @@
...
+      `,
+    template: (uri, intlService) => `
+    <div 
+      property="say:hasPart" 
+      typeof="https://say.data.gift/ns/Title https://say.data.gift/ns/ArticleContainer" 
+      resource="${uri}"
+    >
+      <h4 property="say:heading">
+        <span property="eli:number" datatype="xsd:string">
+          <span class="mark-highlight-manual">Voer inhoud in</span>
```
abeforgit, line 35: We're inserting a textual placeholder where the number should go...

(same issue in all templates)

abeforgit (changes requested):
Works for the most part, but it does introduce a significant performance penalty in even very small documents, to the point where I don't think we can use it like this.

For me the performance does not block a merge though, but before we switch to this version of the plugin in prod I think we do need a serious rework

There's also some issues with the cursor after inserting/moving, but I'm pretty sure that's due to the wonky placeholder implementation we have atm

addon/utils/article-structure-plugin/constants.ts
abeforgit: disregard, misunderstood the logic, and I accidentally had local changes while testing

abeforgit (approved):

PR #34, opened 2022-12-14
Add and convert citation plugin

abeforgit:
differences: 

you can now have multiple citations on one line
you can configure where the plugin triggers by providing a function. It will receive the current state and the datastore, and it should return in which ranges the plugin should trigger.

In order to hook everything up, there is a setupCitationPlugin function available which takes in the config and returns the bundle of things that are needed for the plugin to function

It is now also a keyless plugin, which means multiple instances of the same plugin could be added without issue.

the highlight only triggers after 3 characters of searchtext. This is a userfacing change, but I think it actually makes a lot of sense

- feature(citation): convert citation plugin to typescript
- fix(citation): properly hookup components, plugin and schema
- fix(citation): limit decorating to correct triple pattern
- feat(citation): hook up citation search and input triggers


elpoelma (changes requested):
Seems to work well overall, it just seems that the context card only shows up after typing 3 or more characters after the besluit/decreet keyword, is this behaviour expected?

abeforgit:
yep, that's how the regex is set up

elpoelma (approved):

PR #43, opened 2023-01-05
Feature/template nodes

lagartoverde:


elpoelma (changes requested):
Good start, I just have some remarks:
- When parsing the nodes in `parseDOM`, it can be useful to detect both the prefixed and full versions of the RDFa. This can be done using the namespace util functions and the namespaces defined in the `constants` file.
- Currently there seems to be an issue with the numbering when inserting new articles (with the besluit-plugin). We can look into making the besluit-plugin completely obsolete and implementing the article management by the article-structure plugin.

PR #44, opened 2023-01-05
Rework of date plugin with custom date NodeSpec

elpoelma:


lagartoverde (changes requested):
Works good, I think we are missing 2 things, first of all the date gets setted to the actual time it's now, we should set the hour, minutes, seconds and ms to 0 if we are not specifing them in the modal. Second we have to modify the variable plugin to insert this new node :)

elpoelma:
> Works good, I think we are missing 2 things, first of all the date gets setted to the actual time it's now, we should set the hour, minutes, seconds and ms to 0 if we are not specifing them in the modal. Second we have to modify the variable plugin to insert this new node :)

Ok I will be looking into reworking the insert-variable plugin configuration in order to support this.

addon/plugins/rdfa-date-plugin/nodes/date.ts
```diff
@@ -0,0 +1,81 @@
...
+    },
+  },
+  selectable: true,
+  toDOM: (node) => {
+    const { value, onlyDate } = node.attrs;
+    const datatype = onlyDate ? XSD('date') : XSD('dateTime');
+    const humanReadableDate = value
+      ? formatDate(new Date(value), onlyDate)
+      : onlyDate
+      ? 'Voeg datum in'
```
abeforgit, line 26: Would be good to i18n this

addon/components/rdfa-date-plugin/insert.ts
```diff
@@ -16,29 +16,12 @@ export default class RdfaDatePluginInsertComponent extends Component<Args> {
...
     this.controller.withTransaction((tr) => {
       return tr.replaceSelectionWith(
-        this.schema.text('${date and time}', [
-          this.schema.mark('inline_rdfa', {
-            datatype: 'xsd:dateTime',
-            property: 'ext:content',
-          }),
-        ]),
-        false
+        this.schema.node('date', {
```
abeforgit, line 22: Would be nice to autoselect the inserted date here I think

abeforgit (changes requested):
Note I think we have to add format support before we can release this to prod, will make a jira story

abeforgit (approved):

PR #45, opened 2023-01-10
Rework table of contents plugin

elpoelma:
This PR includes several changes for the table-of-contents plugin:
- Configs are no longer RDF-based, but are now prosemirror-nodetype based. For each node/structure that needs to be shown in the table of contents, it needs to be included in the config.
- Configs are no longer serialized, this seemed to be no longer necessary, instead you can configure the node-spec and node-view in the schema with a specific configuration.
- By default, the plugin uses the `textContent` property of a node in order to represent it in the table of contents. If you want to override this behaviour, you may specify a `outlineText` property on the node-specs that need to be picked up. `outlineText` has the type `(node: PNode) => string`.

### DIFF
--- a/addon/components/table-of-contents-plugin/card.ts
+++ b/addon/components/table-of-contents-plugin/card.ts
@@ -1,16 +1,9 @@
 import { action } from '@ember/object';
 import Component from '@glimmer/component';
-import {
-  TableOfContentsConfig,
-  TABLE_OF_CONTENTS_DEFAULT_CONFIG,
-} from '@lblod/ember-rdfa-editor-lblod-plugins/plugins/table-of-contents-plugin/utils/constants';
 import { ProseController } from '@lblod/ember-rdfa-editor/core/prosemirror';
 
 type Args = {
   controller: ProseController;
-  widgetArgs: {
-    config: TableOfContentsConfig;
-  };
 };
 
 export default class TableOfContentsCardComponent extends Component<Args> {
@@ -33,10 +26,6 @@ export default class TableOfContentsCardComponent extends Component<Args> {
     return result;
   }
 
-  get config() {
-    return this.args.widgetArgs.config ?? TABLE_OF_CONTENTS_DEFAULT_CONFIG;
-  }
-
   @action
   toggle() {
     if (this.tableOfContentsRange) {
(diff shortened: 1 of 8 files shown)


abeforgit (changes requested):
Overal idea is solid, but I have some small remarks
- any particular reason the config is string-based? seems we could avoid some parsing by just passing in the nodespecs directly, or as a list of schema => spec[] functions
- I'm not sure providing a default config adds much value here, since it comes with quite a lot of assumptions about your general schema setup
- why is the config a list of lists? (genuine question, not saying that's a bad thing
- maybe `path` isn't the best name for a config that denotes hierarchy, maybe something like `structureHierarchy`?
- there's some console.log statements floating around still

elpoelma:
> Overal idea is solid, but I have some small remarks
> 
>     * any particular reason the config is string-based? seems we could avoid some parsing by just passing in the nodespecs directly, or as a list of schema => spec[] functions
> 
>     * I'm not sure providing a default config adds much value here, since it comes with quite a lot of assumptions about your general schema setup
> 
>     * why is the config a list of lists? (genuine question, not saying that's a bad thing
> 
>     * maybe `path` isn't the best name for a config that denotes hierarchy, maybe something like `structureHierarchy`?
> 
>     * there's some console.log statements floating around still

* The reason why the config is string based, is that I wanted to support regulatory expressions.
* A default config is indeed a bit unnecessary, I will remove it.
* The config is actually a list of nodeHierarchies => these are the possible hierarchies the table of contents can pick up.

abeforgit:
afaict, the config is actually a list of a list of hierarchies, no? its a `{path: string[]}[]`

elpoelma:
> afaict, the config is actually a list of a list of hierarchies, no? its a `{path: string[]}[]`

indeed

abeforgit:
ah I think I was misunderstanding the point of the hierarchy, I thought it was similar to the hierarchy of the structure plugin, e.g. title > section > chapter, etc

but looking at the default config that's not how it works it seems?

elpoelma:
> ah I think I was misunderstanding the point of the hierarchy, I thought it was similar to the hierarchy of the structure plugin, e.g. title > section > chapter, etc
> 
> but looking at the default config that's not how it works it seems?

No the idea is that e.g. the table-of-contents plugin detects the path/hierarchy `chapter` > `structure_header` and then eventually sets the table-of-contents entry as the content of that `structure_header`. So essentially add all headers of all chapter to the TOC.

PR #46, opened 2023-01-10
Update insert-structure command so it looks forward for the next best position to insert

elpoelma:
When inserting a structure and it cant be inserted into the current container, the insert-command now looks forward into the document to find a container it can be inserted in and inserts it at the end of this container.


lagartoverde (approved):
Works good

PR #48, opened 2023-01-10
Roadsign regulation plugin rework

elpoelma:
This PR does two things:
- it uses the new decision articles and the insert-structure command from the article-structure plugin
- it drops the dependency from the datastore.


abeforgit (approved):
some issues with editing values of location variables, but I don't think that's related to this plugin

PR #51, opened 2023-01-12
Custom roadsign regulation node

elpoelma:
This PR adds a custom roadsign regulation node. This ensures that the variable plugins can easily lookup the zonality for a location variable.


addon/plugins/roadsign-regulation-plugin/nodes.ts
```diff
@@ -0,0 +1,87 @@
...
+        },
+      ],
+      ['div', { property: DCT('description') }, 0],
+    ];
+  },
+  parseDOM: [
+    {
+      tag: 'div',
+      getAttrs(node: HTMLElement) {
+        if (
```
abeforgit, line 59: only worry I have here is that we are throwing out any unknown rdfa properties, like rel,rev, etc
probably won't be a practical issue but just in case?

abeforgit (changes requested):

abeforgit (approved):

PR #53, opened 2023-01-12
Feature/date plugin formats

lagartoverde:


elpoelma (changes requested):
Looks good and works well overall! :smile: Maybe we can just select one of the options by default?

lagartoverde:
The default option is not showing because of a problem with the AuRadioButton, I will try to fix it

elpoelma (approved):
Looks good to me now!

PR #62, opened 2023-01-18
feat(citation): make citation plugin datastore-independent

abeforgit:
Adds a new way of configuring the citation plugin activation ranges, by specifiying a function that returns a set of nodetypes in which the plugin should trigger. This calculation is efficient thanks to the use of a `changedDescendants` utility as demonstrated in the [prosemirror tables plugin](https://github.com/ProseMirror/prosemirror-tables/blob/8ec1e96ae994c1585b07b1ca9dc3292f63e868e0/src/fixtables.ts#L24), found [in this discussion thread](https://discuss.prosemirror.net/t/efficiently-finding-changed-nodes/4280)

The old way of directly specifying the ranges is still supported, but no longer provides the datastore as a second argument, which is the reason for the `breaking` label. This is actually just a more sensible design anyway, removing the implicit dependency of the citation plugin on the datastore. Now the host app can simply calculate the required ranges itself, and can choose to do that using the datastore or not.

With this, the option of specifying nodetypes is actually simply a convenience wrapper, since the same behavior could also be implemented in the host app by calculating the relevant ranges.

In addition, there is now a `motivation` nodespec included for our current usecase, and the dummy app is setup to trigger the citation inside any motivation blocks.

### DIFF
--- a/addon/plugins/citation-plugin/index.ts
+++ b/addon/plugins/citation-plugin/index.ts
@@ -5,20 +5,17 @@ import {
   EditorStateConfig,
   InlineDecorationSpec,
   MarkSpec,
+  NodeType,
   PNode,
   ProsePlugin,
   Schema,
   WidgetSpec,
 } from '@lblod/ember-rdfa-editor';
-import {
-  datastoreKey,
-  ProseStore,
-} from '@lblod/ember-rdfa-editor/plugins/datastore';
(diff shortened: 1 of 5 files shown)


elpoelma (commented):
Works well and the code looks sane, there are just some console.logs still lying around.

PR #63, opened 2023-01-19
Preserve date format across reloads

elpoelma:
This PR ensures that the format of a date is written out in the 'data-format' attribute.


lagartoverde (approved):
Yep, my bad

PR #73, opened 2023-01-23
Make besluit title optional

elpoelma:
This PR does the following things:
- The title of a besluit is now optional.
- Removal of redundant features of the besluit-plugin.
- Modernization of `insertTitle` command of besluit-plugin.


lagartoverde (approved):
Looks good and works as expected :)

PR #74, opened 2023-01-24
Feature/recreate uuids on paste

lagartoverde:
As Elena said, this can be added on the ember-rdfa-editor instead.
Basically recreates the uuids based on 2 properties:
```
recreateUri: true,
uriAttributes: ['mappingResource'],
```
This PR also adds this properties to the variables as discussed in the jira issue.
If you agree with the API we can also add them to other nodes we know they need new uuids like besluit, articles, article structures....


elpoelma (changes requested):
Code looks sane overall. Could you add this to a PR to the editor with the necessary changes? The code also still contains some console logs.

addon/plugins/variable-plugin/nodes.ts
```diff
@@ -20,6 +20,8 @@ const emberNodeConfig: EmberNodeConfig = {
   group: 'inline',
   content: 'inline*',
   atom: true,
+  recreateUri: true,
+  uriAttributes: ['variableInstance'],
```
abeforgit, line 24: does this work? Is variableInstance a URI?

PR #75, opened 2023-01-24
Article-structure-plugin: structure insertion improvements

elpoelma:
This PR includes two changes:
- the insert-structure command now also looks backwards for a container to insert into if no other insertion containers are found
- addition of the optional `limitTo` property to a structure-spec: this property defines whether the insertion range of a structure type should be constrained to another node-type. E.g. the insertion of `besluit-article` nodes should be constrained to the current besluit. You don't want the possibility to occur that when inserting a `besluit-article` it is inserted in another besluit than the current one.

### DIFF
--- a/addon/plugins/article-structure-plugin/commands/insert-structure.ts
+++ b/addon/plugins/article-structure-plugin/commands/insert-structure.ts
@@ -31,6 +31,7 @@ const insertStructure = (
       selection,
       nodeType: schema.nodes[structureSpec.name],
       schema,
+      limitTo: structureSpec.limitTo,
     });
     if (!insertionRange) {
       return false;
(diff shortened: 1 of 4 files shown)


addon/plugins/article-structure-plugin/commands/insert-structure.ts
```diff
@@ -80,21 +85,26 @@ function findInsertionRange(args: {
...
     }
   }
-  const nextContainerRange = findNodes(
-    doc,
-    selection.from,
-    true,
-    false,
-    ({ from }) => {
+  const limitContainer = limitTo
+    ? findParentNodeOfType(schema.nodes['limitTo'])(selection)
```
lagartoverde, line 89: I think this limitTo has to be without quotes right?

lagartoverde (changes requested):
I think I may found a bug

addon/plugins/article-structure-plugin/commands/insert-structure.ts
elpoelma: you are right!

PR #84, opened 2023-01-27
Improve performance of article-structure plugin.

elpoelma:
This improves the performance of the plugin, as this PR ensures that the move/insertion possibility is not checked every state change.

To be tested in combination with https://github.com/lblod/ember-rdfa-editor/pull/564.

### DIFF
--- a/addon/plugins/article-structure-plugin/commands/insert-structure.ts
+++ b/addon/plugins/article-structure-plugin/commands/insert-structure.ts
@@ -15,16 +15,22 @@ import IntlService from 'ember-intl/services/intl';
 import { findNodes } from '@lblod/ember-rdfa-editor/utils/position-utils';
 import { containsOnlyPlaceholder } from '../utils/structure';
 import { findParentNodeOfType } from '@curvenote/prosemirror-utils';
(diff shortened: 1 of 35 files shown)


addon/plugins/article-structure-plugin/commands/insert-structure.ts
```diff
@@ -108,8 +109,28 @@ function findInsertionRange(args: {
     return false;
   };
   const nextContainerRange =
-    findNodes(doc, selection.from, true, false, filterFunction).next().value ??
-    findNodes(doc, selection.from, true, true, filterFunction).next().value;
+    findNodes({
```
abeforgit, line 112: can we make this configurable from the host app, and default to searching everything (aka the old behavior)?

abeforgit (changes requested):

elpoelma:
@abeforgit @lagartoverde I'm gonna put this on hold now, I'm not fully happy with the interface yet.

abeforgit:
this keeps running our ci for no reason so gonna close it (but ill leave the branch up)

PR #87, opened 2023-01-30
We had a wrong format in the long datetime option

lagartoverde:
Small bug on the default formats that caused some problems on qa

### DIFF
--- a/addon/plugins/rdfa-date-plugin/index.ts
+++ b/addon/plugins/rdfa-date-plugin/index.ts
@@ -11,7 +11,7 @@ export const defaultDateFormats: DateFormat[] = [
     label: 'Long Date',
     key: 'long',
     dateFormat: 'EEEE dd MMMM yyyy',
-    dateTimeFormat: 'EEEE dd MMMM yyyy \'at\' HH:mm',
+    dateTimeFormat: 'PPPPp',
   },
 ];
 


addon/plugins/rdfa-date-plugin/index.ts
```diff
@@ -11,7 +11,7 @@ export const defaultDateFormats: DateFormat[] = [
     label: 'Long Date',
     key: 'long',
     dateFormat: 'EEEE dd MMMM yyyy',
-    dateTimeFormat: 'EEEE dd MMMM yyyy at HH:mm',
+    dateTimeFormat: 'EEEE dd MMMM yyyy \'at\' HH:mm',
```
abeforgit, line 14: we should use `PPPPp` instead, now there's a random `at` that's not localized

abeforgit (changes requested):

addon/plugins/rdfa-date-plugin/index.ts
lagartoverde: true, sorry you told me, it's monday, the problem with PPPPp is that it doesn't give the user a start from where customize it's date, but I think is the best solution

abeforgit: what do you mean with "us date"?

lagartoverde: Like from DD/MM/YYYY hh:mm, the user can easily change it to replace the separators, add new words...., this doesn't happen with the PPPPp but I think is also okay

elpoelma (approved):

PR #90, opened 2023-01-31
show error when the date format is empty

lagartoverde:


elpoelma (changes requested):
A red border around the format input-field is only shown when clicking insert. It also does not prevent insertion of a date. When modifying a date, no error is giving when emptying the date-format field.

PR #94, opened 2023-02-03
feat(dates): add nice error messages to custom format box

abeforgit:
Validate the format string and show what's wrong with it
also fixes https://binnenland.atlassian.net/browse/GN-3997 (prefill the correct format from the selected date)


elpoelma (changes requested):
Looks very good! The error message comprehensive and well displayed. Just a small remark on extracting the format when clicking on a date. If a custom format is active, the custom format radio-control is not selected.

lagartoverde:
I get what you are trying to avoid with the dd instead of DD, but BUT, DD is also day of the year, maybe a warning is fine, but we should let the user insert the date just in case they want to use that. Same with yyyy and YYYY.
I don't know users are weird
Screen Shot 2023-02-03 at 14 24 54
Screen Shot 2023-02-03 at 14 27 01

abeforgit:
I'm not recommending anything, the parsing function literally errors when you use DD: 
https://date-fns.org/v2.29.3/docs/format#exceptions

lagartoverde:
> I'm not recommending anything, the parsing function literally errors when you use DD: https://date-fns.org/v2.29.3/docs/format#exceptions

Oh okay, I didn't see you need to use an especial option to use them, then I will approve the PR

lagartoverde (approved):
Looks good, just fix what elena pointed out, everything else works fine, good job 👍

abeforgit:
> Looks very good! The error message comprehensive and well displayed. Just a small remark on extracting the format when clicking on a date. If a custom format is active, the custom format radio-control is not selected.

This is now fixed, along with a complete rework and a lot of polish.
Refer to commit messages to see what changed

PR #112, opened 2023-02-20
Added a new option to the structures to not support unwrap and made the article paragraphs the first structure to use this option

lagartoverde:
Per Bart request, it doesn't really make sense to unwrap an article paragraph. So I added this new option that allows you to define a structure without the unwrap behaviour


elpoelma (approved):
Seems to work well :smile:. Just a small comment: I would use an attribute named `wrappable` which has a default value of `true`, but that's just my personal opinion ofc.

PR #117, opened 2023-03-03
Chore/plugins documentation

lagartoverde:
I'm very bad at writing both in English and Spanish, so feel free to point out any mistakes you find or if you think something is not very readable.
But I guess overall is better than before

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,7 +123,7 @@ You should be able to add a reference manually by clicking on the `Insert` > `In
 ## import-snippet-plugin
 Plugin allowing importing of external RDFA snippets and inserting it in the document.
 
-The `plugin has a card that needs to be added to the sidebar like
+The plugin has a card that needs to be added to the sidebar:
 ```hbs
   <ImportSnippetPlugin::Card @controller={{this.controller}}/>
 ```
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -148,54 +172,51 @@ The plugin can be configured through the following optional attributes:
...
-    publisher: 'http://data.lblod.info/id/bestuurseenheden/141d9d6b-54af-4d17-b313-8d1c30bc3f5b',
-    defaultEndpoint: 'https://dev.roadsigns.lblod.info/sparql',
-    variableTypes: [
-          'text',
-          'number',
-          'date',
-          'location',
-          'codelist',
+  publisher: 'http://data.lblod.info/id/bestuurseenheden/141d9d6b-54af-4d17-b313-8d1c30bc3f5b',
+  defaultEndpoint: 'https://dev.roadsigns.lblod.info/sparql',
```
elpoelma, line 176: The insert-variable plugin examples seem to be outdated.

README.md
```diff
@@ -206,32 +227,41 @@ When the insert-variable-plugin is enabled, users will have the option to insert
 ## rdfa-date-plugin
```
elpoelma, line 227: Can you add an example for the date-insert widget?

elpoelma (changes requested):
Looks good overall I just have some comments.

README.md
lagartoverde: No, it still works like that, the only change is that you have to pass the config to the card instead of the plugin now

README.md
lagartoverde: yes sorry, I forgot about that part

README.md
elpoelma: Yes but it no longer expects a `template` but rather a `constructor` property.

lagartoverde: Sorry didn't realize that had changed should be fixed now

elpoelma:
Can you additionally add a changelog?

elpoelma (approved):

PR #118, opened 2023-03-06
make citation use the new link node

lagartoverde:


addon/plugins/citation-plugin/index.ts
```diff
@@ -269,14 +275,20 @@ function collectDecorations(
...
+    const resolvedNode = doc.resolve(pos);
     if (
       node.isText &&
       node.text &&
-      !schema.marks.citation.isInSet(node.marks)
+      !(
+        resolvedNode &&
+        resolvedNode.parent &&
+        resolvedNode.parent.type === schema.nodes.link
+      )
```
elpoelma, line 291: In here you could check if the resolvedNode has any ancestor of type link.

elpoelma (changes requested):
Seems to work well. I just have some comments.

addon/plugins/citation-plugin/index.ts
lagartoverde: Like this better? :)

elpoelma: You can also use `findParentNodeOfTypeClosestToPos`, you can apply that function on a resolved position, then you don't need to create a node-selection.

elpoelma (approved):

PR #119, opened 2023-03-08
Find table of contents insertion point

lagartoverde:


elpoelma (changes requested):
Looks good overall, I think assuming that table of contents is a direct child of the document is indeed a good heuristic. FYI, you can use `state.doc.descendants` instead of `state.doc.nodesBetween`, then you don't need to supply the start and end indices. Additionally, there are some typescript issues that need fixing surrounding the `replacePosition` variable.

PR #122, opened 2023-03-20
Remove unused code and plugin dependency from citation-insert component

elpoelma:
This PR ensures the citation-plugin functionality can be used without the citation prosemirror-plugin and regex-based autocomplete feature.

In order to integrate the citation-plugin without the autocomplete functionality, only the `CitationPlugin::CitationInsert` needs to be used.


lagartoverde (approved):
Looks good to me

PR #125, opened 2023-03-23
Change date labels based on designer feedback

lagartoverde:
Saska said that putting the formats in the labels was going to be more clear to the end user

### DIFF
--- a/addon/plugins/rdfa-date-plugin/index.ts
+++ b/addon/plugins/rdfa-date-plugin/index.ts
@@ -12,6 +12,7 @@ export const defaultDateFormats: DateFormat[] = [
 ];
 
 export type DateFormat = {
+  label?: string;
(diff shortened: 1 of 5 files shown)


addon/plugins/rdfa-date-plugin/index.ts
```diff
@@ -1,20 +1,17 @@
...
   {
-    label: 'Long Date',
     key: 'long',
     dateFormat: 'EEEE dd MMMM yyyy',
     dateTimeFormat: 'PPPPp',
   },
 ];
 
 export type DateFormat = {
-  label: string;
```
abeforgit, line 17: I'd like to keep the option though, so instead of just removing support for the label, lets just make it optional

abeforgit (changes requested):

PR #132, opened 2023-04-06
Explain the removal options for structures

abeforgit:
- feat(component): add a hover-tooltip utility component
- fix(i18n): properly initialize ember-intl in the dummy app
- feat(structures): separate the remove card into buttons, and add new translations

fixes https://binnenland.atlassian.net/browse/GN-4202

### DIFF
--- a/addon/components/hover-tooltip.hbs
+++ b/addon/components/hover-tooltip.hbs
@@ -1,4 +1,4 @@
-<Velcro @placement="top" as |velcro|>
+<Velcro @placement={{this.placement}} as |velcro|>
   {{#let velcro.loop as |loop|}}
     {{yield (hash velcroHook=velcro.hook handleHover=(modifier this.hover)) to='hover' }}
     {{#if this.tooltipOpen}}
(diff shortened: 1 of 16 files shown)


abeforgit:
@x-m-el adding you as as reviewer just to get you into the flow of things, have a look around 
we use https://keepachangelog.com/en/1.0.0/ in this project and the editor, but not in other projects (though it would get my vote to use it for all of them)

I tend to use conventional commit messages, but that's not a requirement and I'm the only one in the team that does so, feel free to write commit msgs however you want, but make sure they're informative of course :)

addon/components/hover-tooltip.hbs
```diff
@@ -0,0 +1,8 @@
+<Velcro @placement="top" as |velcro|>
```
x-m-el, line 1: As a getter for `placement` exists (and it returns this.args of hover-tooltip), I'm guessing the point is to be able to override the Velcro placement by adding @replacement when using hover-tooltip.
So I'm guessing this should be 
```suggestion
<Velcro @placement=this.placement as |velcro|>
```

lagartoverde (approved):
Looks good to me, just fix mel comment and you can merge 👍

addon/components/hover-tooltip.hbs
abeforgit: ooh good catch, yeah forgot to wire that up in the template :grimacing:

PR #133, opened 2023-04-12
Fix initialization of default date properties

elpoelma:
This PR ensures that the format and value of a data node are correctly initialized to default values if these are not supplied. This PR is a fix to the date issue occuring in https://binnenland.atlassian.net/browse/GN-4183?atlOrigin=eyJpIjoiMzMwNzZlNjgxMDMyNDllM2E0YzIxMzBmY2I3MzU1ZjgiLCJwIjoiaiJ9.


addon/components/rdfa-date-plugin/date-time-picker.ts
```diff
@@ -7,25 +7,25 @@ import { localCopy } from 'tracked-toolbox';
...
   }
 
   get minutes() {
-    return this.date.getMinutes();
+    return this.date?.getMinutes();
   }
 
   get seconds() {
-    return this.date.getSeconds();
+    return this.date?.getSeconds();
```
x-m-el, line 28: This makes sense, but why is it needed if a default is given everywhere (so date shouldn't be undefined anymore)? Future proofing?

x-m-el (commented):
I'm guessing the issue would still persist if a default is not given in some new code.
Probably this wouldn't get used outside of the already defined date-node and this isn't a problem though.

addon/components/rdfa-date-plugin/date-time-picker.ts
elpoelma: Yes indeed, a default is given everywhere. This is just done for future proofing.

addon/plugins/rdfa-date-plugin/nodes/date.ts
```diff
@@ -16,11 +16,9 @@ const date: (options: DateOptions) => NodeSpec = (options) => {
       mappingResource: {
         default: null,
       },
-      value: {
-        default: new Date().toISOString(),
-      },
+      value: {},
       format: {
-        default: 'dd/MM/yyyy',
+        default: options.formats[0].dateFormat,
```
abeforgit, line 21: Inserting a date in the dummy app is broken, I think it's cause of this line

abeforgit (changes requested):

addon/plugins/rdfa-date-plugin/nodes/date.ts
elpoelma: Should be fixed now.

PR #134, opened 2023-04-21
GN-4114: Insert address

dkozickis:
To run this POC you need to

https://user-images.githubusercontent.com/769698/233649835-91dc3515-68a6-480a-922b-7d9339295f2c.mp4

CleanShot 2023-04-21 at 16 38 18@2x


translations/en-US.yaml
```diff
@@ -146,36 +142,45 @@ date-plugin:
...
   insert-variable: Insert variable
   button: Insert
   enter-variable-value: Select value
 
-
 table-of-contents-plugin:
   title: Table of Contents
   toggle: Show Table of Contents
+
+address-plugin:
```
abeforgit, line 178: please add dutch translations

"Insert address": Adres invoegen
back: Terug
cta (what does cta even mean?): Zoekterm
no-match: Geen resultaten

tests/dummy/app/controllers/besluit-sample.ts
```diff
@@ -153,6 +153,9 @@ export default class BesluitSampleController extends Controller {
       link: {
         interactive: true,
       },
+      address: {
```
abeforgit, line 156: I would assume this is no longer needed now

addon/components/address-plugin/insert.ts
```diff
@@ -0,0 +1,103 @@
...
+
+      this.controller.withTransaction((tr) => {
+        tr.replaceSelectionWith(
+          this.schema.node(
+            'block_rdfa',
+            {
+              typeof: 'https://data.vlaanderen.be/ns/adres#Adres',
+              resource: address.identificator.id,
+            },
+            this.schema.node('paragraph', null, [
```
abeforgit, line 81: bit awkward to have adresses always be a block
also it's incredibly easy to nest adresses, leading to nonsensical rdfa

I know it's a PoC, and the focus was the endpoint logic so I suppose its good enough

abeforgit (changes requested):

addon/components/address-plugin/insert.ts
abeforgit: no then its still a block
we'll likely need a custom type

PR #135, opened 2023-04-21
make static TOC look the same in dynamic content

usrtim:
[screen-recorder-fri-apr-21-2023-16-10-01.webm](https://user-images.githubusercontent.com/115715476/233657849-1a133a2f-8fd3-4c98-af53-8956d8969562.webm)

https://user-images.githubusercontent.com/115715476/233785623-1565aefd-9e2d-49aa-a637-597245c16e21.mp4

DEPENDS ON: https://github.com/lblod/frontend-gelinkt-notuleren/pull/455


addon/components/table-of-contents-plugin/ember-nodes/table-of-contents.ts
```diff
@@ -19,11 +19,38 @@ export default class TableOfContentsComponent extends Component<EmberNodeArgs> {
     return this.args.controller;
   }
 
+  get tableOfContentsRange() {
```
abeforgit, line 22: seems to me you can replace this lookup with `this.args.getPos() + this.args.node.nodeSize`, did you try that?

addon/components/table-of-contents-plugin/ember-nodes/table-of-contents.ts
```diff
@@ -19,11 +19,38 @@ export default class TableOfContentsComponent extends Component<EmberNodeArgs> {
...
+  @action
+  onUpdate() {
+    if (this.tableOfContentsRange) {
+      const { from, to } = this.tableOfContentsRange;
+      this.controller.withTransaction((tr) => {
+        const node = this.args.node;
+        const attrs = node.attrs;
+        const newAttrs = { ...attrs, ...this.outline };
+        const newNode = node.type.create(newAttrs, node.content, node.marks);
+        tr.replaceWith(from, to, newNode);
```
abeforgit, line 42: any reason for not using `setNodeMarkup` here?
https://prosemirror.net/docs/ref/#transform.Transform.setNodeMarkup
looks to me you don't even need to calculate the range then and can just use the `args.getPosition` function

addon/plugins/table-of-contents-plugin/utils/index.ts
```diff
@@ -0,0 +1,43 @@
+type OutlineEntry = {
+  content: string;
+  pos: number;
+  children?: OutlineEntry[];
+};
+
+type LI = (
```
abeforgit, line 7: this type is actually just https://prosemirror.net/docs/ref/#model.DOMOutputSpec so you can import that
but not a bad try :)

addon/components/table-of-contents-plugin/ember-nodes/table-of-contents.hbs
```diff
@@ -1,4 +1,4 @@
-<div class='au-u-background-gray-100 au-u-padding-tiny table-of-contents'>
+<div class='au-u-background-gray-100 au-u-padding-tiny table-of-contents' {{did-insert this.onUpdate}}>
```
abeforgit, line 1: this makes it so the exported version doesn't update until you've toggled the the ToC on and off (this only runs when that div is inserted)

I assume you did this cause we had the update cycle issues when running it on every state change, which is understandable

and its actually not a bad attempt, thing is we don't really need the exported ToC to be up to date until we actually export...

I think we need to take another look at the ToC component to fix this properly, but I won't block this PR for it cause the rest of the logic is decent

abeforgit (changes requested):
we cannot release it like this unfortunately, but it's a fundamental issue with the design of the ToC plugin that we're not gonna get fixed within the scope of the original story.
So I'd say adress the comments related to the actual code, and make this a draft pr. I consider the ticket to be completed, and I'll add a comment on Jira to explain. 

Sorry about the unsatisfying conclusion :/

abeforgit:
so after writing that, I had another thought and I think we can get this working _specifically in GN_
basically, since we control when we save the document there, we can simply do this calculation _just before saving_ 
we'll have to do a search to get the node, but I think it could work as a temporary workaround until we rework the ToC plugin to make this easier

addon/components/table-of-contents-plugin/ember-nodes/table-of-contents.hbs
usrtim: yes u are right. I moved the logic to GN, now It's triggered every time before we save the document

usrtim:
> so after writing that, I had another thought and I think we can get this working _specifically in GN_ basically, since we control when we save the document there, we can simply do this calculation _just before saving_ we'll have to do a search to get the node, but I think it could work as a temporary workaround until we rework the ToC plugin to make this easier
> 
> to do that you'd have to
> 
> * extract the `extractOutline` method into a plain function, and export it
> * move the tableOfContentsRange and the onUpdate logic to GN, and run it when we save the document, right before we convert to HTML

Done this, please have a look also here https://github.com/lblod/frontend-gelinkt-notuleren/pull/455

PR #137, opened 2023-05-03
GN-4216: Enable Firefox cursor fix for variables

dkozickis:
https://binnenland.atlassian.net/browse/GN-4216

### DIFF
--- a/addon/plugins/variable-plugin/nodes.ts
+++ b/addon/plugins/variable-plugin/nodes.ts
@@ -13,9 +13,7 @@ import { v4 as uuidv4 } from 'uuid';
 const CONTENT_SELECTOR = `span[property~='${EXT('content').prefixed}'], 
                           span[property~='${EXT('content').full}']`;
 
-// Need to update EmberNodeConfig to include the needsFFKludge property there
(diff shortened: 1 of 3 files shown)


addon/plugins/variable-plugin/nodes.ts
```diff
@@ -13,7 +13,9 @@ import { v4 as uuidv4 } from 'uuid';
 const CONTENT_SELECTOR = `span[property~='${EXT('content').prefixed}'], 
                           span[property~='${EXT('content').full}']`;
 
-const emberNodeConfig: EmberNodeConfig = {
+// Need to update EmberNodeConfig to include the needsFFKludge property there
+// Or allow to augment it somehow else?
+const emberNodeConfig: EmberNodeConfig & { needsFFKludge: boolean } = {
```
abeforgit, line 18: the `createEmberNodespec` function already passes through extra keys, so we just need to fix the type of EmberNodeConfig by adding:
```ts
  [key: string]: unknown;
```

abeforgit (changes requested):

abeforgit (approved):

PR #138, opened 2023-05-10
GN-4233: Remove structure doesn't always work

dkozickis:
JIRA - https://binnenland.atlassian.net/browse/GN-4233

Reuses the check that was removed while splitting into "Remove structure" and "Remove with content"

### DIFF
--- a/addon/components/article-structure-plugin/structure-card.hbs
+++ b/addon/components/article-structure-plugin/structure-card.hbs
@@ -34,7 +34,7 @@
                   @iconAlignment="left"
                   @skin="link"
                   @alert={{true}}
-                  @disabled={{this.currentStructureType.noUnwrap}}
+                  @disabled={{not this.canRemoveStructure}}
                   aria-describedby="remove-tooltip"
                   {{hover.velcroHook}}
                   {{hover.handleHover}}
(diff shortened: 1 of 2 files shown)


addon/components/article-structure-plugin/structure-card.hbs
```diff
@@ -34,7 +34,7 @@
                   @iconAlignment="left"
                   @skin="link"
                   @alert={{true}}
-                  @disabled={{this.currentStructureType.noUnwrap}}
+                  @disabled={{not this.canRemoveStructure}}
```
lagartoverde, line 37: Why not both? Like `(or this.currentStructureType.noUnwrap (not this.canRemoveStructure))` maybe I'm wrong, but I think that this button should be disabled if the current structure don't allow unwrapping or if it cannot be unwrapped

lagartoverde (changes requested):

lagartoverde (approved):

abeforgit (approved):
This is ok to merge, I do question why our schema doesn't allow a structure to have both articles and substructures as content

It seems deliberate, so you'll notice I left a question for the PMs on the jira ticket. In any case this PR solves a clean bug so we merge

lagartoverde:
> I do question why our schema doesn't allow a structure to have both articles and substructures as content

It was a requirement at first, a structure could have articles only if it had a leaf node, but this has changed I think

PR #139, opened 2023-05-11
[Feature] Improved structured text UI

x-m-el:
Implements the new structure UI that is discussed in [GN-4247](https://binnenland.atlassian.net/browse/GN-4247)
Some comments to hopefully help in working further on this. 
This only works with the changes made in [PR editor #788](https://github.com/lblod/ember-rdfa-editor).
Locally you can test this branch by pulling the latest `ember-rdfa-editor` code (master) and running:
- `npm link` in editor ember-rdfa-editor repo
- `npm link @lblod/ember-rdfa-editor` in ember-rdfa-editor-lblod-plugins (this repo)

### code structure
nodes are defined in `addon/plugins/structured-blocks-plugin/nodes` with `addon/plugins/structured-blocks-plugin/nodes/config.ts` containing the definition of the structure and everything that the nodes have in common. The other files have specific information for the node itself.

Everything uses an ember-node. These ember components that are loaded in are defined in addon/components/structured-blocks-plugin.
The main used component is `structured-block`. 

### UI
The UI is not build with a grid, but instead line per line. Note: a grid like structure might be better. The hard part is structuring in a way that hover and shown-hover things have influence on the correct parts.

Every line is: `space for hovered add button` - `space for prefix (e.g. Hoofdstuk 1.` - `space for input field (currently text field)`  - `space for hovered remove button`.
Empty divs with the same classes are used to create the same indentation. See sub-row that basically implements this structure. With `sub-row` you can add a row that is indented to the left/right.

This definitely needs some rework, but with usage of `sub-row` the overhead is really small and for now it is quite easy to use.

### CSS
AU classes are used as much as possible, but often they need some extra changes to follow the (better) figma example. 
Sometimes a div might contain a class that isn't needed. 

### smelly code
Because of refactoring (and focusing on POC), some code might be unused or have a weird structure. E.g. empty divs (although sometimes these have influence on the UI...)

### Todos
- [ ] check if the design is good enough
- [ ] change the input boxes for title and such to text boxes (multiline) or directly to embedded prosemirror 
  - note that currently the text in input boxes is saved in `attrs`. If using embedded prosemirror, an extra node is needed for the embedded to save its info to. -> definitely possible, but not a trivial change.
- [ ] Text inside structure_content (the box under an article) should be added as prosemirror textnodes, but this is not the case. So it doesn't really get "saved" to prosemirror.
- [x] The grey-box hover works, but the add and remove buttons are always show for the current hovered node and all its children. Some custom hover logic is needed here.

- [ ] **the other views are broken (at this point of testing, `link` plugin also seems to not work anymore). Don't merge this before fixing these influences on the other pages/plugins.**

### possible extras (not limited to this)
- [ ] change the UI properties (`showRemoveBorder`, `addBlockHover`) to be tracked properties in ember component, instead of saving them in `attrs`. Note: simple change, the important part is checking that this doesn't break in edge cases. (see [PR editor #788](https://github.com/lblod/ember-rdfa-editor/pull/788))
- [ ] draggable functionality
- [ ] auto numbering of everything (e.g. Artikel 1., Artikel 2. etc)


abeforgit:
this is running our ci all the time, so ill close the pr but leave up the branch

PR #140, opened 2023-05-15
GN-4140: Endpoint config for CitationPlugin

dkozickis:
https://binnenland.atlassian.net/browse/GN-4140

P.S. I will follow up with a PR in embeddable repo when this PR is approved 🙏


lagartoverde (approved):
Looks good

PR #142, opened 2023-05-23
add label to variables

lagartoverde:
https://binnenland.atlassian.net/browse/GN-4298

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- Add label to variables
+
 ## [7.1.0] - 2023-05-18
 
 ### Added
--- a/addon/components/rdfa-date-plugin/date.hbs
+++ b/addon/components/rdfa-date-plugin/date.hbs
@@ -1,0 +1,8 @@
+<AuPill @icon="pencil" @iconAlignment="right" class="variable" {{on 'click' this.onClick}}>
+  {{@node.attrs.humanReadableDate}}
+  {{#if @node.attrs.label}}
+    <span class="label">
+      ({{@node.attrs.label}})
+    </span>
+  {{/if}}
+</AuPill>
--- a/addon/components/rdfa-date-plugin/date.hbs
+++ b/addon/components/rdfa-date-plugin/date.hbs
@@ -1,5 +1,6 @@
 <AuPill @icon="pencil" @iconAlignment="right" class="variable" {{on 'click' this.onClick}}>
-  {{@node.attrs.humanReadableDate}}
+
+  {{this.humanReadableDate}}
   {{#if @node.attrs.label}}
     <span class="label">
       ({{@node.attrs.label}})
--- a/addon/components/rdfa-date-plugin/date.ts
+++ b/addon/components/rdfa-date-plugin/date.ts
@@ -1,0 +1,33 @@
+import Component from '@glimmer/component';
+import {
+  DecorationSource,
+  NodeSelection,
+  PNode,
+  SayController,
+  SayView,
+} from '@lblod/ember-rdfa-editor';
+import { action } from '@ember/object';
+
+type Args = {
+  getPos: () => number | undefined;
+  node: PNode;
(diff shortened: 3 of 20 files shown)


x-m-el (changes requested):
Some testing that popped up problems:
- I can't edit a label of a variable after it has been inserted (the box for the variable doesn't pop up so I can't edit it at all, except for typing in it). I don't see this mentioned in the design though. So maybe this functionality is not needed. However I do see code for it in `updateVariableLabel` so I'd guess you are supposed to :shrug: 
- `When the user clicks on the pill to fill in the variable, the cursor will set itself in front of the brackets with the label, which is where the user can start typing (unless otherwise stated, like with the number variable)` of the design is not implemented. You can get to this state by inserting a variable and deleting the placeholder text (type something and then delete everything). It becomes really hard to type in the variable now. As the design mentions, clicking on the label-part of the variable should also place the cursor inside the variable text input.
- The variable text takes the same size as the text it surrounds. E.g. adding a variable in a h1 (koptitel 1) heading will create a variable with the label being really big. For me this is not the main problem. The problem is that the variable input (or the text of the variable itself) doesn't scale. so it looks really weird and ugly. Because the input of a variable is an embbeded prosemirror instance (I think?) just fixing the scaling of the variable label is probably the easiest option.


I am not sure how I should test that the labels get "saved" properly (that `toDOM` and `parseDOM` are configured correctly). Looking at the code it looks correct, but is there anyway to actually test this (easily)?

abeforgit:
@x-m-el your last question has a easy answer: you can use the "show export preview" button in the dummy app

x-m-el:
> @x-m-el your last question has a easy answer: you can use the "show export preview" button in the dummy app

And using "customHTML" to import it back to check if parseDOM works. I knew about both, didn't think about using them in sequence to test :)

lagartoverde:
Now depends on https://github.com/lblod/ember-rdfa-editor/pull/801

x-m-el (commented):
1.   A label is still big inside a header/the value is still small inside a header:
afbeelding
2. still can't edit a variable's label once it has been inserted (but maybe this is not needed, as users can just remove and recreate it?)
3. Clicking on a variable label doesn't seem to work perfectly yet. When having empty text like this:
afbeelding
I can only get inside the space to edit the text of the variable by using arrow-keys. Nothing with the mouse seems to put my cursor in the correct place. Clicking on the label itself also doesn't do this. I can select the label text, but it is supposed to instead go inside the place to edit the variable, not select the label's text.

4. label does not show up for a date variable.

x-m-el (changes requested):
see above changes for actual review (making a review here so you can hopefully re-request a review from me after you fixed stuff).

lagartoverde:
I solved the problem with inserting variables in a heading, I'm not really happy with that, but I think is not in the scope of this PR, I will comment it with arne in the standup and will create another ticket for it, as an inner editor should pick all the marks from the outer editor

abeforgit:
> It looks like if you run `npm run lint:fix` locally it will fail. Woodpecker seems to be picking up eslint cache or something and not failing.

note woodpecker runs `lint:js` and `lint:hbs`, not `lint:fix`

x-m-el (changes requested):
afbeelding
Weirdly enough, the value does show big for dates... However, we'll keep that for a different issue: [GN-4347](https://binnenland.atlassian.net/browse/GN-4347)

Can't edit a variable's label after it was created, but that's kinda related to [GN-4346](https://binnenland.atlassian.net/browse/GN-4346). So we can ignore this for now, because it is not yet clear if and how we'd want to do this.

I think that if you import an old document, the labels won't show up. Would make sense for the "default label" to automatically show up by the parsing (so changing `parseDOM` to show a default label if one wasn't specified in attribute `dataLabel`). I am not sure how feasible this is. This will only happen for old documents (new ones will always have the label in the html). So it is maybe a question of how many documents/templates there are that won't have this label because of this.

- [ ] BUG: the date variable is broken :(  The buttons to change the format don't work anymore.

x-m-el (approved):
- bug with dates fixed
- old variables now also have date
- no other issues found: 
  - tested insert variables and dates 
  - tried to break date plugin, but kept working :)
  
Good to merge.

PR #145, opened 2023-06-01
GN-4220: Generic RDFA insert

dkozickis:
First iteration on https://binnenland.atlassian.net/browse/GN-4220 (Generic rdfa variables input plugin), allows to insert RDFA as "plain" HTML.

Exploring folder-tree UI as next step. 

https://github.com/lblod/ember-rdfa-editor-lblod-plugins/assets/769698/7a28aaaf-95a8-416a-b0be-fd5199fa1e11

### DIFF
--- a/addon/components/generic-rdfa-variable/insert-menu.ts
+++ b/addon/components/generic-rdfa-variable/insert-menu.ts
@@ -35,7 +35,15 @@ export default class GenericRdfaVariableInsertMenu extends Component<Args> {
   closeModal() {
     this.modalOpen = false;
 
-    // awaiting next tick, otherwise the editor will not be focused properly
+    // The recommended way to focus the editor is to call `this.controller.focus()` before opening the modal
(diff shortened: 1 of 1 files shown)


lagartoverde (changes requested):
Looks good to me, but I'm having some problems with the placement of the suggestions, they appear on the bottom of the modal for me
Screenshot 2023-06-01 at 15 17 09

addon/components/generic-rdfa-variable/insert-menu.hbs
```diff
@@ -0,0 +1,30 @@
+<Toolbar::Button
+  @disabled={{this.disabled}}
```
elpoelma, line 2: The `disabled` property does not seem to be defined in the backing class.

elpoelma (changes requested):
Seems to work well overall :) Just a small note.

abeforgit (changes requested):
when opening the modal, the codemirror editor is not focused
when closing the modal (in any way), the editor is not refocused

PR #147, opened 2023-06-08
GN-4258: Number variable - support min/max

dkozickis:
https://binnenland.atlassian.net/browse/GN-4258

https://github.com/lblod/ember-rdfa-editor-lblod-plugins/assets/769698/71bcb49b-3b60-4742-a209-c1387ca44421

- Tests (lint) currently failing because of failure in upstream PR #142

### DIFF
--- a/addon/components/variable-plugin/number-settings.hbs
+++ b/addon/components/variable-plugin/number-settings.hbs
@@ -1,0 +1,31 @@
+<div class='number-settings'>
+  <AuFormRow>
+    {{#let (unique-id) as |id|}}
+      <AuLabel for={{id}}>
+        {{t 'variable-plugin.number.minimum'}}
+      </AuLabel>
+      <AuInput
+        id={{id}}
+        @value={{@minimumValue}}
+        placeholder={{t 'variable-plugin.number.minimum-placeholder'}}
+        @width='block'
+        @type='number'
+      />
+    {{/let}}
+  </AuFormRow>
+  <div class='to'><span>{{t 'variable-plugin.number.to'}}</span></div>
+  <AuFormRow>
+    {{#let (unique-id) as |id|}}
+      <AuLabel for={{id}}>
+        {{t 'variable-plugin.number.maximum'}}
+      </AuLabel>
+      <AuInput
+        id={{id}}
+        @value={{@maximumValue}}
+        placeholder={{t 'variable-plugin.number.maximum-placeholder'}}
+        @width='block'
+        @type='number'
+      />
+    {{/let}}
+  </AuFormRow>
+</div>
--- a/addon/plugins/variable-plugin/nodes.ts
+++ b/addon/plugins/variable-plugin/nodes.ts
@@ -9,9 +9,54 @@ import {
   EmberNodeConfig,
 } from '@lblod/ember-rdfa-editor/utils/ember-node';
 import { v4 as uuidv4 } from 'uuid';
+import {
+  MAXIMUM_VALUE_HTML_ATTRIBUTE_KEY,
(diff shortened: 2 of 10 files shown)


addon/components/variable-plugin/number-settings.hbs
```diff
@@ -0,0 +1,31 @@
...
+      <AuInput
+        id={{id}}
+        @value={{@minimumValue}}
+        placeholder={{t 'variable-plugin.number.minimum-placeholder'}}
+        @width='block'
+        @type='number'
+      />
+    {{/let}}
+  </AuFormRow>
+  <div class='to'><span>tot</span></div>
```
lagartoverde, line 16: tot should also be translated :)

addon/plugins/variable-plugin/utils/constants.ts
```diff
@@ -92,23 +94,25 @@ export const DEFAULT_VARIABLE_TYPES: Record<string, VariableType> = {
...
         schema.node('placeholder', {
-          placeholderText: selectedCodelist?.label,
+          placeholderText: codelist?.label,
         })
       );
     },
   },
 };
+
+export const EXTRA_ATTRIBUTES_KEY = 'extraAttributes';
```
lagartoverde, line 118: maybe the extra attributes should be written to the html in the data- attributes

lagartoverde (changes requested):
Looks good to me overall, just left some minor comments, I don't like the name extra attributes but I don't have a better one, I just think we should use the data- atttributes instead of writting it as extraattributes in the exported HTML

x-m-el:
I'm not sure if I'm looking at the code wrong, but why is "EXTRA_ATTRIBUTES_KEY" needed?
I know the whole variable structure is a bit of a mess (see [GN-4346](https://binnenland.atlassian.net/browse/GN-4346) ), and this feels like trying to maybe partially fix this, but it doesn't feel like the correct way.

I think it'd be better to add the min and max as separate attributes (if using `data-` is a convention for this stuff, could be called data-min and data-max).
This will of course only be used by a number variable, but it will just be empty (and not used) by other variables, which I think is fine.

x-m-el (commented):
I'm not sure if I'm looking at the code wrong, but why is "EXTRA_ATTRIBUTES_KEY" needed?
I know the whole variable structure is a bit of a mess (see [GN-4346](https://binnenland.atlassian.net/browse/GN-4346) ), and this feels like trying to maybe partially fix this, but it doesn't feel like the correct way.

I think it'd be better to add the min and max as separate attributes (if using `data-` is a convention for this stuff, could be called `data-min` and `data-max`).
This will of course only be used by a number variable, but it will just be empty (and not used) by other variables, which I think is fine (not the best, but again, this is because it needs to be split up in the future anyway).

lagartoverde:
Looks good to me, but still I would rather it be written as `data-minimumvalue` instead of `minimumvalue`

lagartoverde (approved):

addon/plugins/variable-plugin/nodes.ts
```diff
@@ -107,29 +160,44 @@ const emberNodeConfig: EmberNodeConfig = {
...
-          const datatype = [...node.children]
+
+          const datatype = nodeChildren
             .find((el) => hasRDFaAttribute(el, 'property', EXT('content')))
             ?.getAttribute('datatype');
-          if (!mappingResource) {
-            return false;
-          }
+
+          const label = node.getAttribute('dataLabel');
```
elpoelma, line 199: This should be `node.getAttribute('data-label');`. Alternatively, you can also use `node.dataset.label`

elpoelma (changes requested):
Looks good overall, but still noticed some issues

PR #153, opened 2023-06-12
change variable label to be stores in the data-attributes and solve bug with date not getting a default label

lagartoverde:


x-m-el (approved):
Good catch. I must have looked over the label not showing up anymore :)

PR #157, opened 2023-06-13
[NumberVariable] Input number UI + validation is number and min/max

x-m-el:
**issue**: [GN-4259](https://binnenland.atlassian.net/browse/GN-4259)

### How To Test
- add a number variable with optional min/max value
- click on the number variable node
  - you can't change the number directly. A pop-up appears with an input textbox.
  - an error will show if anything that isn't a number is inputted or if min/max is not followed.
  - if wrong data is inputted, this is not saved to the DOM

### Implementation
- Note that min/max is also implemented (see [GN-4258](https://binnenland.atlassian.net/browse/GN-4258)).
- If more validation is added, this is probably better split into different functions.
- The chose was made to only show one error at once, to avoid overstimulation for the user. However, this does mean a user might only see what they should do after a few tries
  - this problem mostly comes from the min/max not showing anywhere. This is out of scope for this PR, but maybe should be **added as a ticket**? Kinda related to https://binnenland.atlassian.net/browse/GN-4346, where it would be nice if the "template user" view shows some information about a variable.
- I avoided any refactoring to keep this simple but working. This means that the code is not perfect, but just function extraction to make the number ember-node work and some if-cases if it is a number node. Again, would be something to look into in [GN-4346](https://binnenland.atlassian.net/browse/GN-4346)
- When clicking on a number node, I wanted to automatically focus the input box. However, this is not accessible, so I removed that. A user can still get to the input box with keyboard by using tab. They can not leave it though (maybe pressing enter should close the box and set cursor right after the node?)

### DIFF
--- a/addon/plugins/variable-plugin/nodes.ts
+++ b/addon/plugins/variable-plugin/nodes.ts
@@ -15,9 +15,9 @@ import {
   MINIMUM_VALUE_HTML_ATTRIBUTE_KEY,
   MINIMUM_VALUE_PNODE_KEY,
 } from './utils/constants';
-import { PNode } from '@lblod/ember-rdfa-editor';
+import { Attrs, DOMOutputSpec, PNode } from '@lblod/ember-rdfa-editor';
 
-const CONTENT_SELECTOR = `span[property~='${EXT('content').prefixed}'],
+export const CONTENT_SELECTOR = `span[property~='${EXT('content').prefixed}'],
                           span[property~='${EXT('content').full}']`;
 
 export const getHTMLNodeExtraAttributes = ({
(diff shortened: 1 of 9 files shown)


lagartoverde:
Tested but didn't look at the code yet but I found a problem, as you said the input box is not actually focused but it shows as selected so my first instinct was to start typing which caused the number to appear on the document as I replaced the selected number variable for my number.
For an inexperienced user this might look like the behaviour of a placeholder bypassing the validation.

addon/plugins/variable-plugin/nodes.ts
```diff
@@ -58,7 +58,110 @@ export const getPNodeExtraAttributes = ({
   return {};
 };
 
-const emberNodeConfig: EmberNodeConfig = {
+export const parseAttributes = (node: HTMLElement): false | Attrs => {
```
lagartoverde, line 61: With the variable number being a separate node now both variables will have different parsing and while is good so simplify code in a function I don't it's something we can do here, or at least not like this.
The minimun and maximum value should be part only of the number variable for example

lagartoverde (changes requested):

x-m-el:
extra bug found:
image
The pencil icon gets shown above the card. Probably need to increase the z-index of the card (this is done for a link-component too, probably for the same reason)

addon/plugins/variable-plugin/nodes.ts
elpoelma: @x-m-el I think this is a good approach for now, the merge conflicts will indeed be more minimal

elpoelma (changes requested):
Functionality-wise, the number editing works well. The validation and error messages seems to be clear. 
Just some notes on the design:
- The size of the edit number pop-up seems rather large
- We should probably refrain from using the placeholder style inside the variable as users will think they can type directly in the modal

x-m-el:
> Functionality-wise, the number editing works well. The validation and error messages seems to be clear. Just some notes on the design:
> 
>     * The size of the edit number pop-up seems rather large
> 
>     * We should probably refrain from using the placeholder style inside the variable as users will think they can type directly in the modal

- The input box can definitely be smaller now, but will be that large because under it a toggle button with "use full text" will be present:
image
I didn't style the size because in the near future it'll need to be bigger anyway :shrug: 

- Good point about the placeholder. I used it to make clear if nothing was filled in yet (avoiding an empty variable which looks broken). As this wasn't specified in the design I discussed this with Saska to know for any alternatives:
  - She says using the placeholder style is okay. The user expects input a variable, which they can do with the input box popping up. Trying to show it different will be another style which can quickly become chaotic.
  - Personal note about this: I think variables will still see a lot of change and the confusion might be a problem when every variable works just a bit differently. So I'd keep the comment in mind for the future, but follow Saska and leave the placeholder look for now.

x-m-el:
Changes:
- decreased input box padding and make the size static (before the size would change depending on the shown error).
- keep tooltip on the bottom left (instead of moving with bigger numbers :) )
- autofocus the input box when selecting the number variable. Pressing enter defocuses it. 
- fix bug of pencil icon showing above the input box (tooltip)
- make the input an actual number input: It gives some auto validation (can't type non-number things) which is nice, but not sure if this has other side effects we might not want.

Should be ready for re-review

x-m-el:
@lagartoverde do you still want to check this, or do I merge? :)

elpoelma (approved):

lagartoverde:
You can merge if Elena approved

PR #158, opened 2023-06-14
Feature/number to words

lagartoverde:
Creating draft PR mainly so I don't forget, need to base my work on Ruben work, but still has to change a lot leaving it for after that PR is more mature

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## Added
+- Numbers inputted into a number variable are validated on defined min/max and if it is a number
+
 ## [7.2.0] - 2023-06-13
 ### Fixed
 - Bump `@lblod/ember-rdfa-editor` package to fix annotation not present for some structures
--- a/addon/components/variable-number/number.ts
+++ b/addon/components/variable-number/number.ts
@@ -1,0 +1,69 @@
+import Component from '@glimmer/component';
+import {
+  DecorationSource,
+  NodeSelection,
+  PNode,
+  SayController,
+  SayView,
+} from '@lblod/ember-rdfa-editor';
+import { action } from '@ember/object';
+import { tracked } from '@glimmer/tracking';
+import { service } from '@ember/service';
+import intlService from 'ember-intl/services/intl';
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore
+import { localCopy } from 'tracked-toolbox';
+
+type Args = {
+  getPos: () => number | undefined;
+  node: PNode;
+  updateAttribute: (attr: string, value: unknown) => void;
+  controller: SayController;
+  view: SayView;
+  selected: boolean;
+  contentDecorations?: DecorationSource;
+};
+export default class VariableNumberPluginNumberComponent extends Component<Args> {
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-call
+  @localCopy('args.node.attrs.value', '') declare inputNumber: string;
+  @tracked errorMessage = '';
+  @service declare intl: intlService;
+
+  get node() {
+    return this.args.node;
+  }
+  get formattedNumber() {
+    return this.node.attrs.value as string;
+  }
+
+  get selected() {
+    return this.args.selected;
+  }
+
+  @action onInputNumberChange(event: InputEvent) {
+    this.inputNumber = (event.target as HTMLInputElement).value;
+    this.validateAndSave();
+  }
+
+  validateAndSave() {
+    if (!Number(this.inputNumber)) {
+      this.errorMessage = this.intl.t('variable.number.error-not-number');
+      return;
+    }
+
+    this.errorMessage = '';
+    this.args.updateAttribute('value', this.inputNumber);
+  }
+
+  @action
+  setSelected() {
+    const tr = this.args.controller.activeEditorState.tr;
+    tr.setSelection(
+      NodeSelection.create(
+        this.args.controller.activeEditorState.doc,
+        this.args.getPos() as number
+      )
+    );
+    this.args.controller.activeEditorView.dispatch(tr);
+  }
+}
--- a/addon/components/variable-number/number.ts
+++ b/addon/components/variable-number/number.ts
@@ -45,11 +45,11 @@ export default class VariableNumberPluginNumberComponent extends Component<Args>
   }
 
   get minValue() {
-    return this.node.attrs[MINIMUM_VALUE_PNODE_KEY];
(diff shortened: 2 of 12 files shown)


x-m-el:
as mentioned in Standup (just to not forget when reviewing :) ):
the actual value of the number should be added as rdfa somehow (with "content"?), as the written way is not useful for extracting information.

addon/components/variable-number/number.ts
```diff
@@ -44,7 +44,18 @@ export default class VariableNumberPluginNumberComponent extends Component<Args>
...
+
+    if (
+      !writtenNumber ||
+      Number.isNaN(Number(this.node.attrs.value)) ||
+      this.node.attrs.value === null ||
+      this.node.attrs.value === ''
+    ) {
+      return this.node.attrs.value as string;
+    } else {
+      return n2words(Number(this.node.attrs.value), { lang: 'nl' });
```
x-m-el, line 57: I personally like the number more with "includeOptionalAnd" set to true, but I asked in jira if Bart knows of any preference, because both are officially correct "Belgian Dutch".

An actual config for this could later be added together with the idea of a "document language" (which Arne has talked about). But for now we need to at least decide on a default.


There is also the option "noHundredPairs" (note: the wiki docs are wrong, this is the correct spelling of the option :) ), but I can't find a source online what the correct spelling would be... I also asked this to Bart.

addon/components/variable-number/number.ts
```diff
@@ -11,14 +11,13 @@ import { action } from '@ember/object';
...
 import intlService from 'ember-intl/services/intl';
-// eslint-disable-next-line @typescript-eslint/ban-ts-comment
-// @ts-ignore
 import { localCopy } from 'tracked-toolbox';
 import {
   MAXIMUM_VALUE_PNODE_KEY,
   MINIMUM_VALUE_PNODE_KEY,
 } from '@lblod/ember-rdfa-editor-lblod-plugins/plugins/variable-plugin/utils/constants';
 import { isBlank } from '@ember/utils';
+import n2words from 'n2words';
```
x-m-el, line 20: You can specify only importing one language ([wiki-link](https://github.com/forzagreen/n2words/wiki/Importing-only-specific-languages), but I see it gives some typescript errors, so if you also couldn't get it to work, this is fine (I feel like the problem is with the library somewhere).
But I wanted to at least mention it :)

addon/components/variable-number/number.ts
```diff
@@ -30,8 +29,9 @@ type Args = {
   contentDecorations?: DecorationSource;
 };
 export default class VariableNumberPluginNumberComponent extends Component<Args> {
-  // eslint-disable-next-line @typescript-eslint/no-unsafe-call
   @localCopy('args.node.attrs.value', '') declare inputNumber: string;
+  @localCopy('args.node.attrs.writtenNumber', '')
```
x-m-el, line 33: the second parameter is a default value if `writtenNumber` is undefined. If you do want to give it a default, this should be a boolean, not an empty string.

addon/components/variable-number/number.ts
```diff
@@ -44,7 +44,18 @@ export default class VariableNumberPluginNumberComponent extends Component<Args>
     return this.args.node;
   }
   get formattedNumber() {
-    return this.node.attrs.value as string;
+    const writtenNumber = Boolean(this.writtenNumber);
```
x-m-el, line 47: why is this needed? Maybe because the default value isn't a boolean? (see other comment)

addon/components/variable-number/number.ts
x-m-el: Beside that: I'm guessing writtenNumber is a localCopy because AuToggleSwitch is a two-way bind (as it uses <Input> internally): https://github.com/appuniversum/ember-appuniversum/issues/375 (ugh). 
Just note that normally this wouldn't be needed. But for now it does :(

addon/components/variable-number/number.ts
x-m-el, line 57: This would only be needed if invalid html is imported (as normally attrs.value is always a valid number because of the validation). But as you validate the input with contentToDom, this isn't really needed.

Besides that, you'd want to split this if statement anyway, with an if/else using only `!writtenNumber`.

addon/plugins/variable-plugin/nodes.ts
```diff
@@ -52,12 +58,40 @@ export const getPNodeExtraAttributes = ({
...
+  type,
+  node,
+}: {
+  content: string;
+  type: string;
+  node: PNode;
+}) => {
+  if (type === 'number') {
+    if (
+      node.attrs.writtenNumber ||
```
x-m-el, line 81: as you defined a constant for the name of this attribute (WRITTEN_NUMBER_PNODE_KEY), you should use it here too

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 ### Added
 - Numbers inputted into a number variable are validated on defined min/max and if it is a number
+- Add a way for the user to insert numbers as words
```
x-m-el, line 11:
```suggestion
- Add toggle for the user to show number variable as words
```
I think this is a bit more explicit :)

x-m-el (changes requested):
Some bugs:
- [ ] inserting html with number not written in full, will automatically toggle it to written in full
- [ ] might be bug from other code. also hard to explain, so let me know if you don't understand:
  - insert a number variable with max 10
  - type "102"
  - will show 10 in document, 102 in input field and an error
  - export the html
  - import that same html
  - the "102" and error will still be present, even though it shouldn't.
  This *only* happens if the nodes are the same. Probably because prosemirror reuses the node and it doesn't get reset. So only happens if you set almost the same content as is already present. But I still think this could be a bug.
  There is also the question if this is bad: should a user lose their previous (wrong) number when e.g. pressing save and the document gets reset? I think it should just for consistency, but I'm not sure.
  
Some of my other comments might fix these issues.

addon/components/variable-number/number.ts
x-m-el: Bart has answered (and probably the final answer):
- No "en" in words
- don't care about the hundredPairs

So both config options can stay their default (which is false)

addon/components/variable-number/number.ts
```diff
@@ -44,7 +44,20 @@ export default class VariableNumberPluginNumberComponent extends Component<Args>
     return this.args.node;
   }
   get formattedNumber() {
-    return this.node.attrs.value as string;
+    const writtenNumber = this.writtenNumber;
```
x-m-el, line 47: I think this is not needed (probably forgot to delete/change it?)
It would make sense to add `const value = this.node.attrs.value as string` to make code more readable though.

lagartoverde:
Created the issue https://binnenland.atlassian.net/browse/GN-4400 for the problem with the toggle

x-m-el (approved):

PR #162, opened 2023-06-16
GN-4323: Extract query tools

dkozickis:
### Overview

Extract SPARQL Query utils so they can be reused

### DIFF
--- a/addon/utils/sparql-helpers.ts
+++ b/addon/utils/sparql-helpers.ts
@@ -1,23 +1,23 @@
 import * as RDF from '@rdfjs/types';
 import { optionMapOr } from '@lblod/ember-rdfa-editor-lblod-plugins/utils/option';
 
-export type QueryResult<Binding = Record<string, RDF.Term>> = {
+export interface QueryResult<Binding = Record<string, RDF.Term>> {
   results: {
     bindings: Binding[];
   };
-};
+}
 
-type QueryConfig = {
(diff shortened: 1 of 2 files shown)


addon/utils/sparql-helpers.ts
```diff
@@ -1,13 +1,25 @@
...
+  query: string;
+  endpoint: string;
+  abortSignal?: AbortSignal;
+};
+
+export async function executeQuery<Binding = Record<string, RDF.Term>>({
+  query,
+  endpoint,
+  abortSignal,
+}: { query: string } & QueryConfig) {
```
abeforgit, line 20: looks like you forgot to change this after you extracted the type :)

addon/utils/sparql-helpers.ts
```diff
@@ -1,13 +1,25 @@
 import * as RDF from '@rdfjs/types';
+import { optionMapOr } from '@lblod/ember-rdfa-editor-lblod-plugins/utils/option';
 
-export type QueryResult = {
+export type QueryResult<Binding = Record<string, RDF.Term>> = {
```
abeforgit, line 4: this is so minor that I don't expect you to change it, but I tend to prefer `interface` when possible, for no real reason other than consistency

abeforgit (changes requested):

abeforgit (approved):

PR #163, opened 2023-06-16
GN-4323: Query the backend for fragments

dkozickis:
### Overview

This PR queries the SPARQL backend for fragments and just displays their title in the Search Modal.

### Connected issues and PRs:

### Setup

* Create some snippets there
* Go to `https://localhost:4300`, observe same as in video

https://github.com/lblod/ember-rdfa-editor-lblod-plugins/assets/769698/e58bcc57-1fbc-4810-8275-cc8c1e846f84

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Numbers inputted into a number variable are validated on defined min/max and if it is a number
 - Fragment insert modal
+- Query backend for fragments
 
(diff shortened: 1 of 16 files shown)


abeforgit (changes requested):
cant get this to work unfortunately
If I run the count query directly on my backend, it returns the expected results, so the data is there.
But when I use the plugin, it always returns 0. 
Might be something with my setup, but can you doublecheck?

EDIT: figured it out, I was sending the requests to the identifier (which also has a /sparql route) rather than directly to the database service

abeforgit (approved):

PR #164, opened 2023-06-16
GN-4323: Introduce pagination for Fragments

dkozickis:
### Summary

Extract pagination components from citations plugin, so it can be reused for fragments

### Setup

* Create some snippets there
* Go to `https://localhost:4300`, click "Fragment insert" in the sidebar, observer screenshot

CleanShot 2023-06-16 at 12 40 52@2x


abeforgit (changes requested):
the pagination doesn't work:
the UI is correct, but the request is never refired when the page changes (cause there's nothing that ties the request to the tracked page property)
I suggest using ember-resources to do so

abeforgit (approved):

PR #165, opened 2023-06-16
GN-4323: Insert fragment

dkozickis:
### Overview

This PR completed the 5 part series of PRs for Fragment insert, you can actually insert a fragment with this PR :D 

### Connected issues and PRs:

### Setup

* Create some snippets there
* Go to `https://localhost:4300`, observe same as in video

https://github.com/lblod/ember-rdfa-editor-lblod-plugins/assets/769698/75c64db8-2719-476a-8a49-19b7cd316782


tests/dummy/app/controllers/besluit-sample.ts
```diff
@@ -237,8 +238,16 @@ export default class BesluitSampleController extends Controller {
...
         interactive: true,
       },
-      fragment: {
-        endpoint: 'http://localhost:8890/sparql',
+    };
+  }
+
+  get fragmentConfig(): FragmentPluginConfig {
+    return {
+      endpoint: 'http://localhost:8890/sparql',
```
abeforgit, line 246: rather point this to the dev endpoint for consistency with the other plugins

abeforgit (changes requested):
- fragments in list are editable
- fragment name not shown (this is a design oversight, but still)
- search field unfocuses when results are found (and focuses one of the snippets)
- closing and opening the modal doesn't reset the search, but also doesn't keep the search string in the searchfield
- doesn't show the content if content has structures:
image
and in the modal:
image

abeforgit (changes requested):
- you can still ctrl-z in the fragment
- if the fragment contains a list, you can still edit it
- if the fragment contains a variable pill, you can still edit it

abeforgit:
For convenience, could you please merge this pr stack into 1, it's getting quite annoying to work with

abeforgit (changes requested):
insertion menu is shown twice
image


and yeah with the overlay... you can still tab into the interactive elements...

I also feel like we're making life unnecessarily hard on ourselves... we have to pass through the whole editor config to get it to render out correctly, etc while the thing we're trying to render _is html already_

Can you get Bart and Saska's opinion on whether this is all really necessary? could we get by with the read-only styles we already have for regulations?

image

E.g. this is what the regulation insert modal looks like in GN... and we're essentially doing the same thing here...

tests/dummy/app/controllers/regulatory-statement-sample.ts
```diff
@@ -174,6 +175,20 @@ export default class RegulatoryStatementSampleController extends Controller {
       link: {
         interactive: true,
       },
+      snippet: {
+        endpoint: 'http://localhost/sparql',
+      },
+    };
+  }
+
+  get snippetConfig(): SnippetPluginConfig {
```
abeforgit, line 184: this is now no longer used

abeforgit (changes requested):
looks good now, just make sure to cleanup unused code

sorry for the detour, but we learned some valuable intel about the editor in the meantime

Make sure to update the endpoint when the dev endpoint is updated to work correctly

PR #183, opened 2023-06-26
GN-4365: add optional internationalization support to static version of table of contents

elpoelma:
### Overview
This PR adds internationalization support to the static version of the table-of-contents node. When constructing the table_of_contents nodespec and nodeview, an optional `intl: IntlService` argument can be provided.

Jira: https://binnenland.atlassian.net/browse/GN-4365?atlOrigin=eyJpIjoiZmI3OWExMWQ1Mzc5NGUzN2FiOTA0MjUwMmJmNTQzM2MiLCJwIjoiaiJ9

### How to test/reproduce
Enable the table-of-contents node on the regulatory statements dummy route and check if the static version is correctly internationalized.

### Challenges/uncertainties
- I'm not completely sure about the interface, the main issue I'm trying to solve is giving access to ember-intl to the `toDOM` node-spec method.

### Checks PR readiness
- [x] changelog
- [x] npm lint

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Breaking
+- Expect additional `intl: IntlService` argument for the `table_of_contents` node in order to support an internationalized static version.
+
 ## [8.1.0] - 2023-06-22
 ### Added
 - Numbers inputted into a number variable are validated on defined min/max and if it is a number
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-### Breaking
-- Expect additional `intl: IntlService` argument for the `table_of_contents` node in order to support an internationalized static version.
+### Added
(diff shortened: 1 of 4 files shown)


x-m-el (commented):
-- not review, just a note! --
Maybe I'm missing something, but wouldn't the TOC use the localePlugin from https://github.com/lblod/ember-rdfa-editor/pull/857 to choose what language to show?
This is more general, as the user can choose exactly which text to choose in which context. The main problem I see is the inconsistency: the documentlanguage (via localPlugin) might be different from the language that `intl` will get.

elpoelma:
> -- not review, just a note! -- Maybe I'm missing something, but wouldn't the TOC use the localePlugin from [lblod/ember-rdfa-editor#857](https://github.com/lblod/ember-rdfa-editor/pull/857) to choose what language to show? This is more general, as the user can choose exactly which text to choose in which context. The main problem I see is the inconsistency: the documentlanguage (via localPlugin) might be different from the language that `intl` will get.

Indeed, but I'm also hesitating because of the inconsistency, maybe it's better to just use the language `ember-intl` gets.

elpoelma:
I will update this PR once I finished up the PR for document language support on the editor.

elpoelma:
> this depends on the document language right? if so, you need to bump the peerdep before merging

Indeed, but I'll update this PR once I have a custom serializer working.

PR #186, opened 2023-06-30
GN-4404: Made the number variable also show placeholders

lagartoverde:
### Overview
Reworked the variable plugin to show the placeholder nummer when no number was inserted

[jira ticket](https://binnenland.atlassian.net/browse/GN-4404?atlOrigin=eyJpIjoiZTA0YmM3YTAwNWNhNDA5YTlmZjJiMTNlNTI4ZDEyYzYiLCJwIjoiaiJ9)

### Setup
none

### How to test/reproduce
Insert a number variable, don't fill it and then click on the show html export button, you should see nummer instead of nothing 

### Challenges/uncertainties
Currently only nummer gets written, which is correct for now, but it has to change when the document language gets implemented in https://github.com/lblod/ember-rdfa-editor/pull/872

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/addon/plugins/variable-plugin/nodes.ts
+++ b/addon/plugins/variable-plugin/nodes.ts
@@ -78,15 +78,14 @@ export const contentToDom = ({
   node: PNode;
 }) => {
   if (type === 'number') {
-    if (
-      node.attrs[WRITTEN_NUMBER_PNODE_KEY] ||
-      Number.isNaN(Number(content)) ||
-      content === null ||
-      content === ''
-    ) {
-      return n2words(Number(content), { lang: 'nl' });
+    if (!Number.isNaN(Number(content)) && content !== null && content !== '') {
+      if (node.attrs[WRITTEN_NUMBER_PNODE_KEY]) {
+        return n2words(Number(content), { lang: 'nl' });
+      } else {
+        return content;
+      }
     } else {
-      return content;
+      return 'nummer';
     }
   } else {
     return content;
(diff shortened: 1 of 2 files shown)


addon/plugins/variable-plugin/nodes.ts
```diff
@@ -78,15 +78,14 @@ export const contentToDom = ({
...
-      return n2words(Number(content), { lang: 'nl' });
+    if (!Number.isNaN(Number(content)) && content !== null && content !== '') {
+      if (node.attrs[WRITTEN_NUMBER_PNODE_KEY]) {
+        return n2words(Number(content), { lang: 'nl' });
+      } else {
+        return content;
+      }
     } else {
-      return content;
+      return 'nummer';
```
abeforgit, line 88: something like 'Voeg getal in' would be better

abeforgit (changes requested):
Doesn't quite work consistently
repro in dummy: 
- add number variable (and do nothing else)
- "show export preview" -> placeholder not shown
repro in RB:
(- npm link_
- make new regulation
- add number var (and do nothing else)
- click on "save and publish"

I'm adding the "do nothing else" cause it seems as soon as you do other things it does start working.

elpoelma:
> Doesn't quite work consistently repro in dummy:
> 
>     * add number variable (and do nothing else)
> 
>     * "show export preview" -> placeholder not shown
>       repro in RB:
>       (- npm link_
> 
>     * make new regulation
> 
>     * add number var (and do nothing else)
> 
>     * click on "save and publish"
> 
> 
> I'm adding the "do nothing else" cause it seems as soon as you do other things it does start working.

Should be solved now, the main issue was the distinction of having the value of the content attribute being `null` vs `undefined`

abeforgit (approved):

PR #204, opened 2023-07-11
GN-4395: make `errorMessage` of number input modal reactive to attribute changes

elpoelma:
### Overview
This PR ensures that the error message shown in the number input popup updates automatically if any node attributes change (such as the minimum and maximum value of the number variable). Currently it is not possible to update the min/max value through a UI when editing the variable, but this PR ensures the error message is updated when changing the html of the document manually.

Solves a large part of https://binnenland.atlassian.net/browse/GN-4395?atlOrigin=eyJpIjoiNjdhZGNkZmRjZGQ3NGY2ZWFkNmY5MzNlMjI5OGMzOWEiLCJwIjoiaiJ9

### How to test/reproduce
- Update the minimum or maximum value of a number variable through the `customHTML` modal and see how a possible error message appears (or disappears).

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations


elpoelma:
@lagartoverde tagging you in this to see if you think this is a good solution

x-m-el (approved):
Good change and fix! happy to see the weird bug was because of actual faulty code and not some weird hard-to-debug issue.
Making the error message a getter is more sensible anyway here.

As far as I can tell the connected issue is also solved, except that the input number is not always reset (because localCopy does not change when min/max changes). You could maybe solve this with `@trackedReset`, however, it would reset when pressing "show as words" button too.
Any other solution would be quite overkill for this issue.

In my opinion I'd say:
- good change, approved :+1: 
- ticket can be closed (the important part is fixed)
  - that the "wrong" input number stays visible isn't a big deal and could be seen as a feature anyway. As long as the error message is correct (which you fixed here), I see no problem with it anymore.

elpoelma:
> Good change and fix! happy to see the weird bug was because of actual faulty code and not some weird hard-to-debug issue. Making the error message a getter is more sensible anyway here.
> 
> As far as I can tell the connected issue is also solved, except that the input number is not always reset (because localCopy does not change when min/max changes). You could maybe solve this with `@trackedReset`, however, it would reset when pressing "show as words" button too. Any other solution would be quite overkill for this issue.
> 
> In my opinion I'd say:
> 
>     * good change, approved +1
> 
>     * ticket can be closed (the important part is fixed)
>       
>       * that the "wrong" input number stays visible isn't a big deal and could be seen as a feature anyway. As long as the error message is correct (which you fixed here), I see no problem with it anymore.

I also tried it with the reset, but thought it would be confusing to the user when it resets the input when the toggle is clicked.

PR #206, opened 2023-07-11
GN-4418: display number numerically if too big to convert to words

elpoelma:
### Overview
When trying to convert a number which is too big using the `n2words` function, `n2words` throws an error. This caused the editor to crash.
This PR introduces a wrapper around `n2words` which displays a warning if `n2words` throws an error and returns the provided number numerically.
This ensures that numbers which are too big to be represented in words are always displayed numerically.
Additionally, the 'to words' toggle is disabled if it is not possible to convert the current number to words.

### Linked issues
- Jira ticket: https://binnenland.atlassian.net/browse/GN-4418?atlOrigin=eyJpIjoiNmU4MzlkZDVkMzgwNDBlYmI0YTI2NmVkYjljYWIzZTkiLCJwIjoiaiJ9 

### How to test/reproduce
Insert a number and input a high value (above 999 trillion) and select the 'to words' toggle. The number should still be displayed numerically and an error should no longer be thrown.

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations


addon/components/variable-number/number.ts
```diff
@@ -68,6 +70,15 @@ export default class VariableNumberPluginNumberComponent extends Component<Args>
...
   }
 
+  get canConvertToNumber() {
+    try {
+      n2words(Number(this.node.attrs.value), { lang: 'nl' });
+      return true;
+    } catch (e) {
+      return false;
+    }
+  }
```
x-m-el, line 80: I feel like this would fit together in the file with `numberToWords`. That way you can avoid using n2words at all in this file, making it so (in the future), devs will more quickly grab the safer numbertoWords, instead of using n2words because they see it here. And they just kinda fit together: both are custom functions to handle that n2words can error.

addon/components/variable-number/number.ts
```diff
@@ -35,6 +36,7 @@ export default class VariableNumberPluginNumberComponent extends Component<Args>
   @tracked errorMessage = '';
   @service declare intl: intlService;
   cursorPositionKeyDown: number | null = null;
+  @tracked infoTooltipOpen = false;
```
x-m-el, line 39: am I missing something, or is this unused code that should be removed?

x-m-el (changes requested):
Good solution.
Some nitpick:
It seems like the number is still somewhat displayed as a string (with e-notation) because of the `toString` method. This is "fine" because the user chose to display as words, but the button is disabled.

In other words: when writing a big number, the state of the toggle button does change the way the number gets displayed.

In a way, I don't think disabling the button is needed and also solves this issue.
Show as number: still works
show as words: works, but if too big it'll use e-notation
And you can toggle between them even for (too big) numbers.

elpoelma:
> Good solution. Some nitpick: It seems like the number is still somewhat displayed as a string (with e-notation) because of the `toString` method. This is "fine" because the user chose to display as words, but the button is disabled.
> 
> In other words: when writing a big number, the state of the toggle button does change the way the number gets displayed.
> 
> In a way, I don't think disabling the button is needed and also solves this issue. Show as number: still works show as words: works, but if too big it'll use e-notation And you can toggle between them even for (too big) numbers.

You're right, I will remove the code which disables the toggle.

addon/components/variable-number/number.ts
elpoelma: Yes it should be removed indeed :)

x-m-el (approved):
clean :ok_hand: 
Tested and couldn't find any more problems.

PR #208, opened 2023-07-12
Fix number variable tooltip's css when inside a header

x-m-el:
### Overview
Before the tooltip box would look like this:
afbeelding
With too much padding between the input and other buttons.

Now it looks like this:
afbeelding

I added some 5px padding too, as before everything sticked together a bit too much.

No ticket was created, as I saw this when testing a different PR.

### How to test/reproduce
Add a number variable and you'll see the difference.


elpoelma (approved):
Nice change! :smile:

PR #209, opened 2023-07-12
GN-4386: validate minValue smaller than maxValue for number variable creation

x-m-el:
### Overview
Before any number could be inputted for the minimum and maximum value of a number. This means that you could insert a number variable that would not expect any number. Now, this is validated, so the minimum value is always smaller than the max value.

https://binnenland.atlassian.net/browse/GN-4386

### How to test/reproduce
- Insert a number variable
  - error shown of min = 2 and max = 1 (or any invalid combination).

### Challenges/uncertainties
I fixed the ticket issue and tried to add a bit more validation. However, not everything works as well as it should. I avoided fixing too much because I didn't want to refactor everything. When https://binnenland.atlassian.net/browse/GN-4346 is done, this should definitely be revisited (together with all other variable code/validation, as it is a bit chaotic now).

- the min/max are now part of a class, so these can be tracked. They are also only added as attributes if they are valid numbers. Mostly avoids adding empty string as an attribute. This wouldn't error, but is info that isn't needed.
  - This is what ember guide adviced to support tracking (otherwise gotta deal with getters/setters...).
  - Just using plain variables instead of adding them in an "extraAttributes" class could work too. That will probably be more sensible when a refactor is done and these variables are more separate (in their own component), instead of all in one big variable-input component.
 
**Notes that are part of the code but not ticket:**

- The values min/max are two-way binded with the input (in `number-settings.hbs`). This two-way bind does not update for non-numbers (input type is set to `number`). You can type text in the input field, but it will not actually update the value (our `minimumValue` and `maximumValue`)
  - this could be confusing if someone inserts `o` instead of `0` and doesn't understand why the validation doesn't work
  - Either way we should avoid two-way binding, but that was a bigger change than desired for this PR ticket fix.
  - probably want some validation about the input being a number too.
- note that you can still have invalid min/max combination by just inserting html of them where it is invalid. I decided this is okay because: 
  - this _should_ be purely a user error (as our editor should not be able to create this state)
  - the number variable input component can handle this, it will just not accept any number. The error will kinda show the problem ("number has to be between 1 and 0").
  - => A solution could be to add another type of error message to the number variable component like "min is bigger than max. There is no possible number to input".


elpoelma (approved):
Seems to work well :). We can indeed look into this again when doing the variable refactor.

PR #215, opened 2023-07-18
update ember-rdfa-editor from 3.10 to 4.0

x-m-el:
### Overview
bumps ember-rdfa-editor to the next "major" release. Includes some nice fixes for ember-nodes, so handy for the current tickets that are being worked on.

### How to test/reproduce
Check the release notes: https://github.com/lblod/ember-rdfa-editor/releases/tag/v4.0.0

Two most important aspects:
- stopEvent and ignoreMutation work differently for ember-nodes. This *should* just work better and not break anything. But to "test" this, you should see if plugins using ember-nodes still work as expected. (think of key-bindings and the likes)
- doc nodespec needs a `toDom` function. This is done by using `docWithConfig`. Without this, error will happen.

### Challenges/uncertainties
I haven't done bumping of dependency before, so don't make the assumption this is the correct way to do it. Please actually review :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,9 +12,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - When a number is too big to be converted to words, display it numerically.
+- Can only insert a number variable with a minimum that is smaller than its maximum
 ### Changed
 - Make `errorMessage` of number input modal reactive to attribute changes
+- Number variable input box has a cleaner UI by adjusting the top margins.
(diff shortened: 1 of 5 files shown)


elpoelma (changes requested):
Can you add a changelog entry?
Additionally: it is now better to use `initialize` instead of `setHTMLContent` as it also parses the lang attribute.

x-m-el:
It seems this update was stopped in https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/196
I don't see anything else (plugin-wise) that would need to be fixed before updating though.

What was the reason for the initial revert? Is this actually fixed now?

elpoelma (approved):

PR #218, opened 2023-07-20
GN-4346: major variable rework

elpoelma:
### Overview
This PR includes a major rework of the variable plugin.
It includes the following breaking changes:
- The generic `variable` node-spec has been changed in favor for domain-specific variable node-specs: `text_variable`, `number`, `date`, `location` and `codelist`. Each of these node-specs have their own corresponding ember-node node-views.
- The `insert-variable-card` component has been reworked. It now merely includes a variable-type selector and is responsible for displaying the correct component used to insert a variable of the selected type. You can configure the `insert-variable-card` component by providing it with an array of `VariableConfig` objects. Each `VariableConfig` object contains the label of the variable which should be displayed in the dropdown, and the path to the insert-component which should be displayed when that variable-type is selected.
   - E.g. an example configuration can be:
     ```js
     [
       {
          label: 'text',
          component: {
            path: 'variable-plugin/text/insert',
          },
        },
        {
          label: 'number',
          component: {
            path: 'variable-plugin/number/insert',
          },
        },
        {
          label: 'date',
          component: {
            path: 'variable-plugin/date/insert',
          },
        },
      ]
      ```
- The `template-variable-card` component has been removed in favor for two seperate `codelist` and `location` edit components. These components contain some code-duplication, but are less cluttered and more clear IMO.
- Removal of `DEFAULT_VARIABLE_TYPES`. The insertion of the variable types are now handled by the insert components of the different variable types.

Jira: https://binnenland.atlassian.net/browse/GN-4346?atlOrigin=eyJpIjoiNzVmYmViODFiMTM0NDcyNWFlNTdiZWY4NmU2YzlmZTAiLCJwIjoiaiJ9

### How to test/reproduce
The UI/UX of inserting and editing variables remains the same as before.

### Challenges/uncertainties
- By splitting up the large insert and edit components into different small ones, a fair bit of code duplication has been introduced. While I generally prefer to avoid code duplication as much as possible, I think it is, in this case, clearer to leave the components split up.

### Checks PR readiness
- [x] changelog
- [x] npm lint

### TODO:
- [x] Update README


addon/plugins/variable-plugin/variables/location.ts
```diff
@@ -0,0 +1,117 @@
...
+    ...sourceSpan,
+    [
+      'span',
+      {
+        property: EXT('content').prefixed,
+      },
+      0,
+    ],
+  ];
+};
```
x-m-el, line 89: there are some parse-helpers (in attribute-parsers.ts) which are really nice and make everything so much clearer. I do miss that it doesn't have its opposite in the same file to use in toDOM. It would be nice to also decrease the complexity of toDOM in some way, because the "core" is quite simple, but it looks very complex with all the lists (ugh).

I am not sure if you can easily make helpers that do what is needed though. But it feels worth it to maybe try :D

addon/components/variable-plugin/insert-variable-card.ts
```diff
@@ -1,193 +1,38 @@
...
-      selection.node.type === this.args.controller.schema.nodes.variable
-    ) {
-      return false;
-    } else {
-      const variable = findParentNodeOfType(
-        this.args.controller.schema.nodes.variable,
-      )(selection);
-      return !variable;
-    }
+    return !this.controller.inEmbeddedView;
```
x-m-el, line 36: This makes it so the card will still show when selecting the date or number variable (as they don't use the embedded view).
This is not bad perse: it will just replace the number or date variable when inserting (so a logical functionality).

We can leave it as is, but it is still a bit inconsistent.
Potential solutions: also hide when it is a node selection or even specifically a variable selection (but that's probably harder?)

README.md
```diff
@@ -573,41 +465,169 @@ For very custom setups, the plugin might be unable to find your scrollContainer
 },
 ```
 
+## variable-plugin
 
-## template-variable-plugin
+Editor plugin which provides node-specs and components which allow you to insert and edit different types of variables in a document. The plugin provides the following variable types, but this can be extended:
```
x-m-el, line 470: did I miss something or is there no explanation what means "can be extended"?

x-m-el (changes requested):
I didn't look at all the code to see if there might be subtle differences, but could a bit of duplication for codelist and location be avoided with adding some functions under "utils"? e.g. insert in edit, wrapVariableInHighlight. Don't want to overdo it, but if they are "general" functions that really do the same, I still think it could make sense? 

Putting toDom and parseDOM in a variable is a big plus, makes everything so much more clearer. Amazing how some code moving can help a ton too :)

For the rest some comments which are mostly just opinions. Feel free to ignore them (even if the reason is "too much work to fix" :) ). Already big improvement that opens the way for increasing code quality.
I also tested it as much as I could and didn't find any problems.

addon/components/variable-plugin/insert-variable-card.ts
elpoelma: I can do a check if the selected node has group `variable` for example?

README.md
elpoelma: Indeed, I rephrased it a bit to make this clearer.

elpoelma:
> I didn't look at all the code to see if there might be subtle differences, but could a bit of duplication for codelist and location be avoided with adding some functions under "utils"? e.g. insert in edit, wrapVariableInHighlight. Don't want to overdo it, but if they are "general" functions that really do the same, I still think it could make sense?
> 
> Putting toDom and parseDOM in a variable is a big plus, makes everything so much more clearer. Amazing how some code moving can help a ton too :)
> 
> For the rest some comments which are mostly just opinions. Feel free to ignore them (even if the reason is "too much work to fix" :) ). Already big improvement that opens the way for increasing code quality. I also tested it as much as I could and didn't find any problems.

I moved some common functions for the codelist and location variable types to a utils file.

addon/plugins/variable-plugin/variables/location.ts
elpoelma: Yes I know what you mean. I don't know if it can be improved in a much clearer way like I did in the parseDOM function though :(

addon/components/variable-plugin/insert-variable-card.ts
x-m-el: Yes I honestly think that would be a needed improvement, because then it is based on the structure/content instead of implementation details (using embedded editor or not).

addon/plugins/variable-plugin/variables/location.ts
x-m-el: Could do something like
```
typeSpan(type) { return ['span', { property: DCT('type').prefixed, content: type }] }
contentSpan(content?) {return [
      'span',
      {
        property: EXT('content').prefixed,
      },
      content ? content : 0,
    ],
    
```

It feels a bit over-abstraction, but does make it easier to read `toDOM` and gives a duality between parseDOM and toDOM in functions. So idk :shrug: Would only make it clear if it is a good idea with bigger and harder to understand parse/toDOM combinations.

I don't know if that's better though and let you decide. I'm fine with leaving how it is too :)

x-m-el: I think the main reason I'd find it easier to read is probably that we don't allow long lines, so even a simple span is over 4 lines now, while with the helpers it would be just one line and much shorter.
But that's abstracting because of our choice of formatting configuration. So I'm not gonna say that's a good reason to do so :p

x-m-el (approved):
rereviewed. Looks good. Check out my two comments.
Didn't retest it all though.

addon/components/variable-plugin/insert-variable-card.ts
elpoelma: Updated it :)

addon/plugins/variable-plugin/variables/location.ts
elpoelma: I added some helper functions, I think it should be more readable now :)

PR #219, opened 2023-07-24
Update Dockerfile

nvdk:
### Overview
updates the docker build to serve static files

### How to test/reproduce
`docker build` the image and run it

### Challenges/uncertainties
it could be there was a specific reason for using ember serve, typically if the image is used to mount sources and run those. In that case, please close the PR and add this info to the readme.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+- Update docker build to serve static files
+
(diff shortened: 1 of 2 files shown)


nvdk:
please squash the PR when merging

elpoelma:
> ### Overview
> updates the docker build to serve static files
> 
> ### How to test/reproduce
> `docker build` the image and run it
> 
> ### Challenges/uncertainties
> it could be there was a specific reason for using ember serve, typically if the image is used to mount sources and run those. In that case, please close the PR and add this info to the readme.

There are not any sources which are mounted.

elpoelma (approved):
Works

PR #220, opened 2023-07-24
add `template comments` support

x-m-el:
### Overview
This adds "template comments": a block of text that can be inserted "anywhere" to give some extra explanation how to fill in a document.
This text is only informational and should be hidden at publishing (_currently not implemented yet_).

Also bumps editor to [4.2.0](https://github.com/lblod/ember-rdfa-editor/releases/tag/v4.2.0). This does not include any breaking changes, but does include changes needed for this PR.

- [template comments GN-4264](https://binnenland.atlassian.net/browse/GN-4264)
- [ember-rdfa-editor#908 PR for embedded-editor changes](https://github.com/lblod/ember-rdfa-editor/pull/908)

### Setup
Need https://github.com/lblod/ember-rdfa-editor/pull/908
- use `npm link` in `ember-rdfa-editor`
- use `npm link @lblod/ember-rdfa-editor` in this repo
- run `ember ts:precompile' in `ember-rdfa-editor` to compile the types and avoid those giving errors.

OR, now that [ember-rdfa-editor 4.2.0](https://github.com/lblod/ember-rdfa-editor/releases/tag/v4.2.0) is released:
- no setup needed - 
### How to test/reproduce
- Go to Regulatory Statements "tab"
- insert a template comment with the button under "insert"
- you can type in the template comment and do everything specified in the issue:

see [comment on design](https://binnenland.atlassian.net/browse/GN-4277?focusedCommentId=79273) what should specifically be implemented: 
- indenting
- numbering
- bullet points
- bold text formatting 

On the right side buttons appear to move and delete the template comment.

### Challenges/uncertainties
~~Some known bugs. Specific of embedded editor are written in [#908 PR](https://github.com/lblod/ember-rdfa-editor/pull/908)~~ => these were a problem with the configuration of the schema in the plugins repo specifically. This has been fixed.

Other problems:

<details>
<summary>Details</summary>

- The list plugin has problems.
  1. create a template comment
  2. type some text
  3. try to make it a list item
  4. an error will occur that an undefined value got unwrapped. This is in the `dispatchInner` of the embedded-editor
  Note that it does work if you copy a list item directly as the first item. It seems like the actual wrapping logic of pressing the list-item button is the problem.

- ~~undo seems to not work correctly anymore (also a problem with transactions?)~~ For some reason [this commit](https://github.com/lblod/ember-rdfa-editor/pull/908/commits/008dcd53b49305271004b0e197ff918b20813eed) fixes the problem
- Italicalizing is not needed. The button (CTRL+I) is disabled, but pressing the button still adds it.
  - Easiest solution is probably: add a "non-italic-paragraph" node to use as content for the template comment

</details>

Problems have been fixed by changing from embedded-editor (which should only be used for inline content) to just using content logic of ember node.
- lists now work as expected
- italize has been "disabled" by using a paragraphWithConfig

#### Move buttons
These have been implemented in an easy way to just move between nodes, as it wasn't really specified in the design and "moving up or down" in general is kind of a hard thing to specify (and probably not desired for a comment?).
The implementation seems to not work for nested nodes because `nodeBefore` and `nodeAfter` don't return anything. A better implementation might be to use nodeBefore if possible, and otherwise change to findInAncestor (like the structures-plugin) to find a fitting (allowed) node to insert.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,12 @@
 
 # broccoli-debug
 /DEBUG/
+
+# idea editor
+.idea/.gitignore
+.idea/vcs.xml
+
+# VScode & extensions settings
+.vscode/extensions.json
+.vscode/settings.json
+.favorites.json
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- Add template-comments plugin
+  - A plugin to insert and edit template comments (as a "block")
+  - can be inserted anywhere
+  - have a specific rdfa-attribute so they can be removed at publishing
+
 ## [9.0.1] - 2023-07-24
 
 ### Changed
--- a/README.md
+++ b/README.md
@@ -640,7 +640,7 @@ Logic to insert a template comment is added with
 <TemplateCommentsPlugin::Insert @controller={{this.controller}}/>
 ```
 
-A button to remove it when selected can be shown with
+Buttons to remove and move it when selected can be shown with
 ```hbs
 <TemplateCommentsPlugin::EditCard @controller={{this.controller}}/>
 ```
--- a/README.md
+++ b/README.md
@@ -32,6 +32,7 @@ This addon contains the following editor plugins:
 * [template-variable-plugin](#template-variable-plugin)
 * [validation-plugin](#validation-plugin)
 * [address-plugin](#address-plugin)
+* [template-comments-plugin](#template-comments-plugin)
 
 You can configure your editor like this:
 ```hbs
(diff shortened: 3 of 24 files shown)


x-m-el:
still todo: add documentation to readme :)

x-m-el:
converted to draft :(
Issues: 
- node does not get "highlighted" when clicked (seems like the css is not added for some reason)
- can write outside of the node by adding a very long text

Will fix those first.

elpoelma (commented):
Works really nicely :smile: . Just a suggestion: maybe the cursor should change when hovering over the unselectable parts of the template comment?

elpoelma:
Also noticed it is not really possible to select the node itself, but that is not really a problem IMO.

x-m-el:
> Also noticed it is not really possible to select the node itself, but that is not really a problem IMO.

Yes I thought about that too: it really depends if you want to limit deletion with only the button, or also with selecting + pressing delete. With the button being the same as that of a structure element, it might make sense to keep the same logic (which is only able to delete with the button).
But like we discussed, that was probably just to reuse UI, not the actual logic, so :shrug:

x-m-el:
cursor is now normal if not on top of the text. You can still click it to get the cursor inside.
Insertion positions are also validated better now. 
I think this might be in a state that it can be finished. 

If merged, I'll make a note of the choices made on the ticket:
- insertion and moving is as global as possible, but never inbetween text (always between "blocks")
- question if putting the buttons close to the node instead of in the side wouldn't be better.
- okay if only deletion possible with button.

x-m-el:
I think some linting fails because woodpecker doesn't know about the necessity of the PR of ember-rdfa-editor. Any way to "fix" this, or do we just ignore those errors when merging (if they don't show locally)?

elpoelma:
> I think some linting fails because woodpecker doesn't know about the necessity of the PR of ember-rdfa-editor. Any way to "fix" this, or do we just ignore those errors when merging (if they don't show locally)?

Wouldn't it be better to first release the editor with the needed changes and update this PR?

x-m-el:
> Wouldn't it be better to first release the editor with the needed changes and update this PR?

Yes, obvious answer on a stupid question :sweat_smile: . Need to indeed make sure this PR has the new ember-rdfa-editor version, which means we need to wait for the other PR to be released first.

PR #226, opened 2023-07-27
GN-4370: Allow block content in `article_paragraph`

dkozickis:
### Overview

Allows to insert `inline*` content in `article_paragraph`, that helps insert `table` among other stuff.

https://binnenland.atlassian.net/browse/GN-4370

### Setup

1. Checkout
2. `npm run start`

### How to test/reproduce

Observe that it is not possible to reproduce issue with inserting table in article paragraph described in https://binnenland.atlassian.net/browse/GN-4370, this does not stop from inserting table into other article nodes, the fix is implemented in `embed-rdfa-editor` MR https://github.com/lblod/ember-rdfa-editor/pull/911

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Fixed
+- Allow `block*` content in `article_paragraph`
 
 ## [9.0.2] - 2023-07-28
 ### Dependencies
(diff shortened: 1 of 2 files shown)


elpoelma (changes requested):
Shouldn't this be `block*` as the table is a block node and paragraphs are too?

elpoelma:
> > Shouldn't this be `block*` as the table is a block node and paragraphs are too?
> 
> The "usual" paragraph is `inline*` 🤔 - https://github.com/lblod/ember-rdfa-editor/blob/b477fdba7bc5aaf08777515465f285066e0e5367/addon/nodes/paragraph.ts#L15

yeah but this is not a paragraph in the sense of a `p` tag. It is just called a `paragraph` but should allow any content.

x-m-el:
This is probably a valid solution, but the places where this paragraph is supported is artikel, which has as content `'(block|article_paragraph)+'`.
Which says to put block things before/after paragraphs.

But tbh, it makes more sense to put everything inside article_paragraphs (because that's the reason this structure exists???), so allowing block content in it feels like a good change.

elpoelma (approved):
Looks good!

PR #227, opened 2023-07-28
GN-4323: Snippet insertion fix

dkozickis:
All snippets are essentially full documents, we cannot just insert one `doc` into another `doc`. Did not see that when just trying to insert "bold" text, but started failing on bigger structures.

### DIFF
--- a/addon/components/snippet-plugin/snippet-insert.ts
+++ b/addon/components/snippet-plugin/snippet-insert.ts
@@ -36,15 +36,27 @@ export default class SnippetInsertComponent extends Component<Args> {
   @action
   onInsert(content: string) {
     const domParser = new DOMParser();
+    const parsed = domParser.parseFromString(content, 'text/html').body;
+    const documentDiv = parsed.querySelector('div[data-say-document="true"]');
 
     this.closeModal();
 
-    this.controller.withTransaction((tr) => {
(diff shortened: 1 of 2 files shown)


addon/components/snippet-plugin/snippet-insert.ts
```diff
@@ -36,15 +36,30 @@ export default class SnippetInsertComponent extends Component<Args> {
...
-        ProseParser.fromSchema(this.controller.schema).parse(
-          domParser.parseFromString(content, 'text/html'),
+    if (documentDiv) {
+      return this.controller.withTransaction((tr) =>
+        tr.replaceSelectionWith(
+          ProseParser.fromSchema(this.controller.schema).parse(documentDiv, {
+            preserveWhitespace: true,
+            topNode: this.controller.schema.nodes.doc.create({
+              lang: documentDiv.getAttribute('lang') ?? undefined,
+            }),
```
elpoelma, line 51: This is not necessary as we're are not parsing the document node (we want to prevent it in this case)

elpoelma (changes requested):
Looks good, we can still clean this up later by publishing the api in the editor. Just a comment.

elpoelma (approved):

PR #229, opened 2023-07-30
GN-4442: Use new findNodePosUp/Down functionality to move Template Comments over whole document

x-m-el:
### Overview
Uses `findNodePosUp/Down` to move template comments in a more clean (and general) way. The old implementation barely works.

- https://binnenland.atlassian.net/browse/GN-4442
- need https://github.com/lblod/ember-rdfa-editor/pull/939

### Setup
- checkout https://github.com/lblod/ember-rdfa-editor/pull/939
- `npm link` in editor-repo
- `npm link @lblod/ember-rdfa-editor` in plugins repo
- might need to do `ember ts:precompile` for types

### How to test/reproduce
- include a template comment in some structure in Regulatory Statements part.
- add some structured data (title, article, paragraphs ...)
- click on the template comment and move it up and down, see that this is always possible and it does not skip valid positions.
- It will also not move inside another template comment (even though, content-wise, this would be valid).

- It should never split any nodes
- Should keep the selection inside the template comment intact

### Challenges/uncertainties
note that this PR was changed (algorithm moved to editor-repo). Hence why some of the comments might not fit.


x-m-el:
> Just a mention, have you looked at [lblod/ember-rdfa-editor@`8ba1d46`/addon/utils/_private/position-utils.ts#L88](https://github.com/lblod/ember-rdfa-editor/blob/8ba1d46224c6a4fcd8408d2358051589a43d1135/addon/utils/_private/position-utils.ts#L88) at all? It seems to at least be in the same ballpark, might not be exactly the same but still worth a look

I did look through the utils but misinterpreted what findNodes did so didn't realize it did similar to what I need (cause it would indeed have been a good starting point :upside_down_face: ). I'm also not sure if it will find them in the order that is needed for moving a node (but now looking through the code itself I think it does?). Hard to be certain at a first glance because the code is quite complex to fully interpret.

I'll test to see if it might be what I need with a good-build filter. :)

abeforgit:
dont forget to bump editor in this pr when that pr is merged

x-m-el:
Changelog was merged wrong. Fixed in commit to master: [16b4e288c496bfd2eca0db7861648af0789c98a6](https://github.com/lblod/ember-rdfa-editor-lblod-plugins/commit/16b4e288c496bfd2eca0db7861648af0789c98a6)

PR #230, opened 2023-07-31
GN-4444: Insert snippet first node correctly

dkozickis:
First node of the snippet still was not inserted correctly.

Inserting bunch of `Node`s with `replaceSelectionWith` is worse than inserting `Slice` with `replaceSelection`. As also evident, by the fact that `replaceSelection` is used in other places, not `replaceSelectionWith`.

https://binnenland.atlassian.net/browse/GN-4444

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 ### Dependencies
 - Bumps `@embroider/test-setup` from 1.8.3 to 3.0.1
+- Bumps `@types/ember__application` from 4.0.4 to 4.0.6
+- Bumps `webpack` from 5.76.0 to 5.88.2
+- Bumps `@types/ember__component` from 4.0.11 to 4.0.14
 ### Fixed
 - Allow `block*` content in `article_paragraph`
 - Move buttons for template comments are blue
--- a/package-lock.json
+++ b/package-lock.json
@@ -4039,16 +4039,18 @@
       }
     },
     "node_modules/@jridgewell/source-map": {
-      "version": "0.3.2",
-      "license": "MIT",
+      "version": "0.3.5",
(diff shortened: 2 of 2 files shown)


elpoelma:
I'm not really able to reproduce the issue where the first node is not inserted correctly. Could you give me an example?

abeforgit:
> Inserting bunch of Nodes with replaceSelectionWith is worse than inserting Slice with replaceSelection.

in what way? are there performance issues or something?

elpoelma:
> Note the first node is not correct
> 
> **Before**
> 
>  CleanShot.2023-08-03.at.18.04.59.mp4 
> **After**
> 
>  CleanShot.2023-08-03.at.18.05.38.mp4

Aha ok I understand I think. In the first video an empty paragraph is inserted before the snippet, and as paragraphs (or other nodes besides a table of contents) are not allowed before document-titles, prosemirror just tries it best with the document-title. In the second there is no such issue, as no empty paragraph is inserted before the document-title.

But anyway, this is normal behaviour. Some nodes included in a snippet document will not be able to be inserted at the location where you want to insert the snippet.

elpoelma:
> > > Note the first node is not correct
> > > **Before**
> > > CleanShot.2023-08-03.at.18.04.59.mp4
> > > **After**
> > > CleanShot.2023-08-03.at.18.05.38.mp4
> > 
> > 
> > Aha ok I understand I think. In the first video an empty paragraph is inserted before the snippet, and as paragraphs (or other nodes besides a table of contents) are not allowed before document-titles, prosemirror just tries it best with the document-title. In the second there is no such issue, as no empty paragraph is inserted before the document-title.
> > But anyway, this is normal behaviour. Some nodes included in a snippet document will not be able to be inserted at the location where you want to insert the snippet.
> 
> But still, I'm inserting a snippet with document title at the start of the document, might as well get it inserted correctly? thinking

I agree.

elpoelma (approved):

PR #236, opened 2023-08-01
Cleanup code for template comments + update readme

x-m-el:
### Overview
There were some functions present that did nothing, as they existed when using embedded-editor, but template comments don't use embedded-editor, but content directly instead.
The readme did not specify how to fix indenting inside template comments. This is added to the readme.

PR where these issues were introduced: https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/220
indentableInsert changed in: https://github.com/lblod/ember-rdfa-editor/pull/923

### How to test/reproduce
check if template comments still work with the removed functions. Or at least check if they were not used.

ADDED:
Check if indenting works as expected.

### Challenges/uncertainties
Indentation was fixed this way to make it easily configurable. However, it might make more sense to create an "indentable" group, that can be added to list items and paragraphs and paragraphWithConfig?
Note: template comment does override the groups of ParagraphWithConfig, so it would have to be added there too. But that is a sensible place to specify it (it is close to the specification of a template node).
=> ADDED specified as a nodespec instead. Nothing has to be changed for template comments, accept for removing `allowedTypes` in the dummy app, as it is not needed anymore.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Changed
+- remove unused code from template comment component
+- remove allowedTypes from indentation menu
 ### Dependencies
 - Bumps `@types/uuid` from 9.0.0 to 9.0.2
 - Bumps `@types/rdf-validate-shacl` from 0.4.0 to 0.4.2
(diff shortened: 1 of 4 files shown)


elpoelma (approved):
Looks good. I like the idea of an `indentable` group, another approach is to introduce a custom `indentable` node-spec property, to not clutter the group property to much.

elpoelma:
> Looks good. You can merge this once a changelog is added :) I like the idea of an `indentable` group, another approach is to introduce a custom `indentable` node-spec property, to not clutter the group property to much.

x-m-el:
@elpoelma  Yes, I like the node-spec way more. Groups are more used for defining the schema, so I agree it would clutter the groups. 
I'll try it out with node-spec and ask for a re-review :+1:

x-m-el:
changed here to use nodespec. A new PR (on ember-rdfa-editor) is now connected.

x-m-el:
this has been closes and the cleanup changes have been moved to https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/245

PR #241, opened 2023-08-02
Set correct group for TemplateCommentParagraph

x-m-el:
### Overview
see most of the explanation in https://github.com/lblod/ember-rdfa-editor/pull/918
This changes the ParagraphWithConfig's group config to be `paragraphGroup`, so it can be included in a list. It also passes a `name` so that the specific paragraph will be recognized in parsing

https://binnenland.atlassian.net/browse/GN-4445
- bug that you could not add lists in template comments (no issue was created for this)>

### Setup
npm link with https://github.com/lblod/ember-rdfa-editor/pull/918

### How to test/reproduce
Add a template comment, see you can add lists.
When reloading, see that paragraphs are not converted to lists, or more importantly, in the structure see that it stays the same (and does not get converted to e.g. a regular paragraph).

### Challenges/uncertainties
BEFORE MERGE: this needs an update to ember-rdfa-editor. So before merging, the other PR needs to be released and the version should be updated here...

Note: linting (precompile) error in CI will be fixed with updating ember-rdfa-editor version

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 ### Fixed
 - Allow `block*` content in `article_paragraph`
+- Add paragraphGroup to paragraphForTemplate so it can be wrapped by list
+- add a name to paragraphForTemplate so it gets parsed as itself 
 
 ## [9.1.1] - 2023-08-02
 ### Fixed
--- a/addon/plugins/template-comments-plugin/node.ts
+++ b/addon/plugins/template-comments-plugin/node.ts
@@ -52,10 +52,12 @@ export const emberNodeConfig: () => EmberNodeConfig = () => {
 export const templateComment = createEmberNodeSpec(emberNodeConfig());
 export const templateCommentView = createEmberNodeView(emberNodeConfig());
 export const templateCommentNodes = {
-  templateComment: templateComment,
   templateCommentParagraph: paragraphWithConfig({
     marks: 'strong underline strikethrough',
-    // don't add to any groups, so it is only allowed for template comment
-    group: '',
+    // don't add to block group, so it is only allowed for template comment
+    // but add to paragraphGroup so lists accept it
+    group: 'paragraphGroup',
+    name: 'templateCommentParagraph',
   }),
+  templateComment: templateComment,
 };
--- a/addon/plugins/template-comments-plugin/node.ts
+++ b/addon/plugins/template-comments-plugin/node.ts
@@ -57,7 +57,7 @@ export const templateCommentNodes = {
     // don't add to block group, so it is only allowed for template comment
     // but add to paragraphGroup so lists accept it
     group: 'paragraphGroup',
-    name: 'templateCommentParagraph',
+    subType: 'templateCommentParagraph',
   }),
   templateComment: templateComment,
 };


elpoelma (changes requested):
Seems to work nicely in combination with the editor PR. One thing I noticed is that it seems to be impossible to go out of a list when creating one in a template comment.

elpoelma:
> Seems to work nicely in combination with the editor PR. One thing I noticed is that it seems to be impossible to go out of a list when creating one in a template comment.

Note: this issue does not occur when using normal paragraphs instead of templateCommentParagraphs in a template-comment. So possibly an issue with a keymap that does not create a `templateCommentParagraph` when entering out of a list.

x-m-el:
Right, I also noticed that bug but forgot to check if it was still present with these changes.
Will check out. Although not sure how to fully fix this without having to add a config everywhere.

x-m-el:
@elpoelma this is not the only issue. As now the lists in a template comment have normal paragraphs, you can italicize those...
So might need to specify a custom list_item too that only accepts templateCommentParagraphs...

That doesn't fix the issue with the enter adding a normal paragraph though.
Seems to be a more difficult problem than first realized.

elpoelma:
> @elpoelma this is not the only issue. As now the lists in a template comment have normal paragraphs, you can italicize those... So might need to specify a custom list_item too that only accepts templateCommentParagraphs...
> 
> That doesn't fix the issue with the enter adding a normal paragraph though. Seems to be a more difficult problem than first realized.

Hmmm quite annoying, I'm starting to think if we cannot just use the normal paragraph in template comments and allow italic. But apply another different style by default to indicate it's a comment (or another type of italic I don't know).

x-m-el:
I guess for this instance it wouldn't be a big problem, except for muddying up the export
(*or* deleting the "invalid" italics inside toDOM could also solve it).
But UI-wise you'd still want to disable the italic button to avoid it not working when expecting it should.

Would be nice to solve it like this because things like buttons should be "fixed" automatically. For this instance we can "hack" it together too, but in other cases it might not be possible.

So trying to see if a solution is possible can be useful. But I also feel like there won't be a clean way :/
Except for an actual fully separate editor (so a custom schema can be used), but that's overkill and not a good idea with undo/redo and such.

elpoelma:
> I guess for this instance it wouldn't be a big problem, except for muddying up the export (_or_ deleting the "invalid" italics inside toDOM could also solve it). But UI-wise you'd still want to disable the italic button to avoid it not working when expecting it should.
> 
> Would be nice to solve it like this because things like buttons should be "fixed" automatically. For this instance we can "hack" it together too, but in other cases it might not be possible.
> 
> So trying to see if a solution is possible can be useful. But I also feel like there won't be a clean way :/ Except for an actual fully separate editor (so a custom schema can be used), but that's overkill and not a good idea with undo/redo and such.

Yeah I don't know either if there is a really clean solution. Was just wondering if enabling italic by default on template comments is really necessary.

x-m-el:
I think i want to try and find a fix (or at least see how convoluted it is).
if there is an okay fix that isnt a big hack, then we know we can do things like this. If we can't find a fix that isn't basically a hack, we know to not allow such features in the future because of constraints with the editor.

I don't feel like just hacking it now and leaving as it is without knowing if we can do something clean or not if something similar comes back.

I do think it is a good idea to remove the italic constraint, use normal paragraphs and deploy that as a working version of template comments and leave this 'research' separately to avoid template comments getting blocked.

elpoelma:
> I think i want to try and find a fix (or at least see how convoluted it is). if there is an okay fix that isnt a big hack, then we know we can do things like this. If we can't find a fix that isn't basically a hack, we know to not allow such features in the future because of constraints with the editor.
> 
> I don't feel like just hacking it now and leaving as it is without knowing if we can do something clean or not if something similar comes back.
> 
> I do think it is a good idea to remove the italic constraint, use normal paragraphs and deploy that as a working version of template comments and leave this 'research' separately to avoid template comments getting blocked.

I agree, I think it's indeed better to find a clean solution. The italic constraint is not super import (I think), so I think we can live without for now.

x-m-el:
The bugs mentioned above are harder to fix than expected. Template comments are instead simplified and this PR is closes for now: https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/245

PR #243, opened 2023-08-03
GN-4436: review min max implementation of number plugin

elpoelma:
### Overview
This PR introduces the usage of the `AuNativeInput` component and one-way-binding in the number variable widgets. 
The number-input, minimum and maximum fields have been updated to use `AuNativeInput` which uses the native `input` element instead of the ember `Input` component used by `AuInput`.

Jira: https://binnenland.atlassian.net/browse/GN-4436?atlOrigin=eyJpIjoiMzI2OWY1YzUzYTk5NDE4MDk1NmMxYzA2ZWI4OGI3NjciLCJwIjoiaiJ9

### How to test/reproduce
- Insert a number and check if input minimum/maximum values works.
- Check if updating a number-variable works.

### Challenges/uncertainties
A note on the usage of `type=number`: while chromium-based browsers support this input type well (they do not allow to input characters outside of numbers), firefox has a bug which allows inputting text in number inputs (see https://bugzilla.mozilla.org/show_bug.cgi?id=1398528). 
The question which stands: do we use inputs with type `text` everywhere (and show a warning to the user when the input contains characters outside of numbers), or do we use the type `number` and ignore the firefox bug?  

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations


x-m-el:
Since when has AuNativeInput been a thing? :O I thought it didn't exist yet (at least it doesn't exist in documentation as far as I can see)

I would also include the following "cleanup":
- use doc-language when the toggle "as words" is set
- Not sure if better (really depends on the new variable code now): might want to avoid having the logic of how to show the numbers in both the component and the node spec.
What I mean is when checking `writtenNumber` and using `numberToWords` if true in both component and toDom. Really the component already "defines" how it should be shown, and should somehow pass that to toDom? (e.g. via an attribute). Hopefully that makes sense.

elpoelma:
> Since when has AuNativeInput been a thing? :O I thought it didn't exist yet (at least it doesn't exist in documentation as far as I can see)
> 
> I would also include the following "cleanup":
> 
> * use doc-language when the toggle "as words" is set
> * Not sure if better (really depends on the new variable code now): might want to avoid having the logic of how to show the numbers in both the component and the node spec.
>   What I mean is when checking `writtenNumber` and using `numberToWords` if true in both component and toDom. Really the component already "defines" how it should be shown, and should somehow pass that to toDom? (e.g. via an attribute). Hopefully that makes sense.

The `AuNativeInput` component is not part of appuniversum, but is something we created and is included in this repo. 

- On the doc-language: I agree, but would do it in a seperate PR on all applicable plugins. (such as the toc-plugin)
- I don't know if I would add an extra attribute containing the written version of the number (as we already have the `writtenNumber` attribute, that seems a bit redundant). I don't think it's too bad to do the conversion twice.

elpoelma:
> > Since when has AuNativeInput been a thing? :O I thought it didn't exist yet (at least it doesn't exist in documentation as far as I can see)
> > I would also include the following "cleanup":
> > 
> > * use doc-language when the toggle "as words" is set
> > * Not sure if better (really depends on the new variable code now): might want to avoid having the logic of how to show the numbers in both the component and the node spec.
> >   What I mean is when checking `writtenNumber` and using `numberToWords` if true in both component and toDom. Really the component already "defines" how it should be shown, and should somehow pass that to toDom? (e.g. via an attribute). Hopefully that makes sense.
> 
> The `AuNativeInput` component is not part of appuniversum, but is something we created and is included in this repo.
> 
> * On the doc-language: I agree, but would do it in a seperate PR on all applicable plugins. (such as the toc-plugin). I can create a seperate ticket for it.
> * I don't know if I would add an extra attribute containing the written version of the number (as we already have the `writtenNumber` attribute, that seems a bit redundant). I don't think it's too bad to do the conversion twice.

I created the following ticket for it: https://binnenland.atlassian.net/browse/GN-4445?atlOrigin=eyJpIjoiNGM1YmZjNDM3OGI2NDY0Y2I0NmRhYjIwZjgyZDc3MzAiLCJwIjoiaiJ9

x-m-el:
For the Firefox bug: I think having some own validation might still be useful.
The bug is not just that you can input text. If you input any non-number input, it will not use it and not clear the input. This also happens with any space in the input.
Can't we still use the input="number", but still validate when submitting if it is a valid number, or autotracked validation? In theory it shouldn't be a problem in Chrome, but will show an error in Firefox if needed.

abeforgit (approved):
see comments, can be done in other pr though, this one is good as is as well

x-m-el:
@elpoelma I would prefer the validation fix in a separate PR (which does link this this one), will make reviewing easier :)

elpoelma:
I also don't know if I'm in favor of converting it to an `input[type='text']` as we then lose the accessibility of the `input[type='number']` fields.

x-m-el:
Right, that bugs makes it annoying to fix, cause I couldn't find a way to get the "real" value either.
There is a polyfill: https://github.com/jonstipe/number-polyfill
But I'm not sure if it works nicely. In a way this bug would also be perfect to fix in AUInput generally, instead of here in a custom way.
Another way is the check the validity with `.checkValidity()` in the element.
So you could still implement the validation with the numbered input, but will need to use Firefox's validation instead. Not the perfect solution, but I think worth to use and should work fairly well?

elpoelma:
> Right, that bugs makes it annoying to fix, cause I couldn't find a way to get the "real" value either. There is a polyfill: https://github.com/jonstipe/number-polyfill But I'm not sure if it works nicely. In a way this bug would also be perfect to fix in AUInput generally, instead of here in a custom way. Another way is the check the validity with `.checkValidity()` in the element. So you could still implement the validation with the numbered input, but will need to use Firefox's validation instead. Not the perfect solution, but I think worth to use and should work fairly well?

The `checkValidity seems interesting`. I will look into it.

elpoelma:
> Right, that bugs makes it annoying to fix, cause I couldn't find a way to get the "real" value either. There is a polyfill: https://github.com/jonstipe/number-polyfill But I'm not sure if it works nicely. In a way this bug would also be perfect to fix in AUInput generally, instead of here in a custom way. Another way is the check the validity with `.checkValidity()` in the element. So you could still implement the validation with the numbered input, but will need to use Firefox's validation instead. Not the perfect solution, but I think worth to use and should work fairly well?

The `checkValidity seems interesting`. I will look into it.

elpoelma:
> Right, that bugs makes it annoying to fix, cause I couldn't find a way to get the "real" value either. There is a polyfill: https://github.com/jonstipe/number-polyfill But I'm not sure if it works nicely. In a way this bug would also be perfect to fix in AUInput generally, instead of here in a custom way. Another way is the check the validity with `.checkValidity()` in the element. So you could still implement the validation with the numbered input, but will need to use Firefox's validation instead. Not the perfect solution, but I think worth to use and should work fairly well?

The `checkValidity seems interesting`. I will look into it.

PR #245, opened 2023-08-07
Simplify template comments to fix bugs: allow any content and any marks of the doc schema

x-m-el:
### Overview
Template comments had some problems by trying to avoid allowing `<em>` marks inside its content. The problem was that lists also needed to be allowed, but lists (and their button functionality) would use normal paragraphs again. 
This PR allows normal paragraphs for template nodes, removing the constraint of avoiding `<em>` marks. It also includes some cleanup cherry-picked from another PR.

- cherrypicked cleanup commits from https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/236
- A discussino of the problems with the custom paragraphs and list can be found in the comments of https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/241
- because this simplifies template comments, no specific ember-rdfa-editor version is needed 
-  related bug with lists in template comments: https://binnenland.atlassian.net/browse/GN-4445

### Challenges/uncertainties/Remarks
- TODO: create a ticket to look at this issue further and see if something like fully custom content in a child block is a thing that is (easily, without hacks) possible. Potentially this can be done with an embedded-editor, if the implementation for that is fixed to work with block content *and* it is fixed so that a custom schema can be provided.
- to partially follow the design, a base `<em>` mark could be added when adding a template comment, but I couldn't get this to work (yet), so that's why it isn't included here. This can always be added later.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Changed
+- remove unused code from template comment component
+- remove allowedTypes from indentation menu
 ### Dependencies
 - Bumps `@embroider/test-setup` from 1.8.3 to 3.0.1
 ### Fixed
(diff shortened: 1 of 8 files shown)


elpoelma (approved):
Works really nicely :) I was wondering if it is possible to pass a list of marks when creating the template node, but I tried to test it and it did not seem to work.

x-m-el:
> Works really nicely :) I was wondering if it is possible to pass a list of marks when creating the template node, but I tried to test it and it did not seem to work.

Yes I tried that too, but didn't work. Documentation doesn't specify well what that parameter does.
Ticket to check it separately: https://binnenland.atlassian.net/browse/GN-4454

PR #255, opened 2023-08-10
Number variable: check validity of min/max input fields

elpoelma:
### Overview
This PR adds functionality which checks the validity of the min/max input fields each time their value changes. It uses the `checkValidity` API for this.

https://binnenland.atlassian.net/browse/GN-4436?atlOrigin=eyJpIjoiOTEyNTdlNWI4OWU5NDIxNjgyZDAzMWMzY2ZkYzBlZTkiLCJwIjoiaiJ9

### How to test/reproduce
- Open the number insert card on firefox
- Input text which is not a number in the minimum/maximum fields
- An error message should show up and the `insert` button should be disabled.

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Introduce internationalization in the table of contents node based on the document language.
 - The whole table of contents node (include its entries) is now exported in its `serialize` method without the need of an `entries` attribute.
 - GN-4461: update readme to specify needed imports for template comment
+- Check validity on number minimum/maximum inputs
 
 ### Changed
 - Use one-way-binding in variable label input
--- a/addon/components/variable-plugin/number/insert.hbs
+++ b/addon/components/variable-plugin/number/insert.hbs
@@ -11,6 +11,7 @@
       <AuNativeInput
         id={{id}}
         value={{this.minimumValue}}
+        {{on 'input' this.updateMinimumValue}}
         {{this.minimumInput}}
         placeholder={{t 'variable-plugin.number.minimum-placeholder'}}
         @width='block'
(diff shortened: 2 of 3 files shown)


x-m-el:
Any reason the modifiers are needed? As far as I can see, `on "input"` still fires every time (and is good enough), even if the value doesn't actually change.

elpoelma:
> Any reason the modifiers are needed? As far as I can see, `on "input"` still fires every time (and is good enough), even if the value doesn't actually change.

Just did it to make sure the initial values are correct (if for some reason they would be prefilled).

x-m-el (changes requested):
Right, that makes sense, but as I can see this case will not happen, except if we start to show the insert card when selecting a number variable, which might well be the case in the future. I don't mind the modifier though, as it is quite easy to read.

However, if the min/max is not a number, the number variable node will not work at all anyway. I wonder if for these cases we actually want to sanitize the attributes in parseDOM, instead of sometimes handling those cases with some UI thing. Although then the user does not know something was left out so both have (dis)advantages, or something more complicated (like some error-messages about cleanups done to the html).

Besides that ramble that is probably not for this PR:
Looks good and seems to work + solves the problem :)

Note about the code:
might be more clear to still keep the ` on "input"` but also have a "initialValidityCheck" modifier? Because if I see correctly the modifier is to make sure the initial check is done, but afterwards it isn't really needed anymore, as it is then just an input event Listener just like the `on` helper would do.

You could also reduce the amount of code by just keeping a `this.validMinMax` boolean and have one modifier then.

x-m-el:
> It would indeed be clearer to leave the `on "input"` modifier as is, I agree with that. On the converting to 1 modifier I'm not sure tho? You still need to be able to know whether both inputs are valid separately I think?

Yes you're right, I was too quick with my suggestion. You need two booleans (one for every field) indeed.

elpoelma:
Should be updated now to use `on "input"` again

x-m-el (approved):
looks good now :)

PR #262, opened 2023-08-16
GN-4438: Grey out Generic RDFA button in embedded view

dkozickis:

https://binnenland.atlassian.net/browse/GN-4438

### Setup

1. Checkout this branch
2. `npm run start`

### How to test/reproduce

Go to http://localhost:4200/regulatory-statement-sample

Observe that when you are in the "embedded view" (e.g. location) the "Insert generic RDFA button" is greyed out

CleanShot 2023-08-16 at 10 35 46@2x

### Challenges/uncertainties

I decided to just check for being in the embedded view, based on the ticket by Ruben 🙇 I expect that user has RDFA nodes in the main editor if he is using the "Insert generic RDFA".

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/addon/components/generic-rdfa-variable/insert-menu.ts
+++ b/addon/components/generic-rdfa-variable/insert-menu.ts
@@ -1,14 +1,14 @@
 import Component from '@glimmer/component';
 import { action } from '@ember/object';
-import { Command, SayController } from '@lblod/ember-rdfa-editor';
+import { SayController } from '@lblod/ember-rdfa-editor';
 import { tracked } from 'tracked-built-ins';
-
-import { DOMParser as ProseParser } from 'prosemirror-model';
 
 import { basicSetup, EditorView } from 'codemirror';
 import { EditorState } from '@codemirror/state';
 import { html } from '@codemirror/lang-html';
 import { tooltips } from '@codemirror/view';
+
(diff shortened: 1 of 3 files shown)


addon/components/generic-rdfa-variable/insert-menu.ts
```diff
@@ -82,16 +86,32 @@ export default class GenericRdfaVariableInsertMenu extends Component<Args> {
...
+    this.args.controller.doCommand(this.insertCommand(editorContent), {
+      view: this.controller.mainEditorView,
+    });
+
+    this.closeModal();
+  }
+
+  insertCommand(editorContent: string): Command {
+    return (state, dispatch) => {
+      if (!this.controller || this.controller.inEmbeddedView) return false;
```
abeforgit, line 98: now obsolete with my other comment, but as a rule of thumb try to avoid using the controller from inside a command function, and only use the state reference you get passed in.
It's unlikely to be a problem in practice, but there could be some funky timing issues

addon/components/generic-rdfa-variable/insert-menu.ts
```diff
@@ -26,6 +26,14 @@ export default class GenericRdfaVariableInsertMenu extends Component<Args> {
     return this.args.controller.schema;
   }
 
+  get canInsert() {
```
abeforgit, line 29: you technically don't need this check anymore now :sweat_smile: but it doesn't hurt

abeforgit (approved):

PR #263, opened 2023-08-16
GN-4263: update address variable according to updated design

elpoelma:
### Overview
This PR attempts to improve the UX and UI when editing an address variable.
It includes the following:
- The UI of the address-edit card has been updated to provide seperate fields for the municipality, streetname, housenumber and busnumber.
  - The municipality can be prefilled with a default value
  - Both the municipality and streetname fields are implemented using `ember-powerselect` and provide auto-completion
  - Both the housenumber and busnumber are simple input fields, each time these are updated, the address is resolved using the address register (https://docs.basisregisters.vlaanderen.be/) (with a delay), to validate if the address exists.

#### How editing an address variable works:
1. First, the user can select a municipality using the muncipality dropdown menu. This field can already be prefilled if desired. The autocomplete of this menu is implemented by querying the location register API (https://geo.api.vlaanderen.be/geolocation). This API accepts a free-form string and can return a list of municipalities based on the provided input. Note: this API will always return the main municipality. If the user inputs a sub-municipality (e.g. Oostakker), they will get prompted with the main municipality (e.g. Gent).
2. Once a municipality has been selected, the user can select a street located in that municipality. The autocomplete of this menu is implemented by quering the address register API (https://api.basisregisters.vlaanderen.be/v2/adresmatch). This API accepts a municipality and streetname, and can return a list of streetnames that match.
3. Once both the municipality and street have been selected, the user can optionally input a housenumber and busnumber. When any of these fields are updated, a request (with a delay) is sent to the address register (https://api.basisregisters.vlaanderen.be/v2/adresmatch) in order to check if the address exists. Based on the response of this request, the user gets to see a success or error message.
4. Once the necessary fields are filled in, the user can insert the new address.

Jira: https://binnenland.atlassian.net/browse/GN-4263?atlOrigin=eyJpIjoiYzQ3NDdlNWViMTQ5NDMwYWE1YTQ0YmJlZjg3MWMzZjUiLCJwIjoiaiJ9

### How to test/reproduce
1. Insert an address variable and select it
2. Follow the steps described above in order to update the address variable
3. When selecting an address variable that already has a value, the edit fields should be prefilled
4. Check the correctness of translations
5. Input a wrong address to check if the error messages work

### Challenges/uncertainties
- At the moment, inclusion of coordinates has been disabled. The address register provides Lambert72 coordinates instead of the WGS84 coordinates provided by the location register. The address register only provides coordinates for full addresses (not for streets without a housenumber)
- When inserting a street (and municipality) without a housenumber, no zipcode is inserted. The address register only provides zipcodes for full addresses.

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### TODO

- [x]  re-enable the coordinates feature

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - GN-4461: update readme to specify needed imports for template comment
 
 ### Changed
+- GN-4263: update address variable edit UI/UX according to updated design
+- Allow the address municipality-edit field to be prefilled
 - Use one-way-binding in variable label input
 - Use one-way-binding in number variable inputs
 
--- a/README.md
+++ b/README.md
@@ -665,8 +665,12 @@ get codelistEditOptions() {
 This addon provides a seperate edit component which allows users to search for an address and update the select address variable. Additionally, they can also choose whether to include the housenumber of an address.
 You can add this edit-component to a template as follows:
 ```hbs
-<VariablePlugin::Address::Edit @controller={{this.controller}}/>
+<VariablePlugin::Address::Edit @controller={{this.controller}} @defaultMuncipality='Antwerpen'/>
 ```
+
+The edit card can be configured with two arguments:
+- An instance of a `SayController` (required)
+- A `defaultMuncipality` which should be used as the default value of the `muncipality` field in the edit-card (optional)
 
 ## validation-plugin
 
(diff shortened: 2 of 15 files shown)


elpoelma:
For the coordinates, we could also fetch the coordinates from the location register in WGS84 format (as we did before).

elpoelma:
Update:
- the coordinates are now fetched in gml format and are stored using geosparql in the rdfa of an address.
- zipcodes have also been added to addresses without housenumbers using the location register API.

abeforgit (approved):
really, really nice job
the only thing that's missing is having it work inside a location variable. unfortunately I suspect that will be the main usecase. 
However I don't consider that blocking for this pr, let's tackle that separately

PR #275, opened 2023-08-31
GN-4266: Referring to published decisions

dkozickis:
### Overview

Allows to refer to published decisions (from Publicatie) using the `citation-plugin` 

https://binnenland.atlassian.net/browse/GN-4266

### Setup
1. Checkout PR
2. `npm run start`

### How to test/reproduce

Repeat after me (might have to insert motivering, or just change `activeInNodeTypes` in `besluit-sample.ts` to include doc node)

To see the "card" easily apply this diff

```diff
Index: addon/components/citation-plugin/citation-card.ts
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/addon/components/citation-plugin/citation-card.ts b/addon/components/citation-plugin/citation-card.ts
--- a/addon/components/citation-plugin/citation-card.ts	(revision f7740bcacdd11a9de25b6e66a4df3057b149a6a5)
+++ b/addon/components/citation-plugin/citation-card.ts	(date 1693566856337)
@@ -66,6 +66,8 @@
   }
 
   get showCard() {
+    return true;
+
     return !this.controller.inEmbeddedView && this.activeDecoration;
   }
 
@@ -202,15 +204,11 @@
   insertDecisionCitation(decision: LegalDocument): void {
     const uri = decision.uri;
     const title = decision.title ?? '';
-    const { from, to } = unwrap(this.activeDecoration);
+
     this.controller.withTransaction(
       (tr: Transaction) =>
         tr
-          .replaceRangeWith(
-            from,
-            to,
-            citedText(this.controller.schema, title, uri),
-          )
+          .replaceRangeWith(0, 0, citedText(this.controller.schema, title, uri))
           .scrollIntoView(),
       { view: this.controller.mainEditorView },
     );

```

https://github.com/lblod/ember-rdfa-editor-lblod-plugins/assets/769698/5a6e8d8c-f979-49ba-92df-459f46f6c7bf

### Challenges/uncertainties

**NB**
Following files do not contain new code, I have just extracted them from `addon/plugins/citation-plugin/utils/vlaamse-codex.ts`, as it was becoming too big

* `addon/plugins/citation-plugin/utils/vlaamse-codex.ts`
* `addon/plugins/citation-plugin/utils/utils.ts`
* `addon/plugins/citation-plugin/utils/types.ts`
* `addon/plugins/citation-plugin/utils/process-match.ts`
* `addon/plugins/citation-plugin/utils/cache.ts`
* `addon/plugins/citation-plugin/utils/article.ts`

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/addon/plugins/citation-plugin/index.ts
+++ b/addon/plugins/citation-plugin/index.ts
@@ -82,9 +82,14 @@ const MINISTERIAL = `ministerieel${NNWS}?besluit`;
 const ENUM_ROYAL = `genummerd${NNWS}?${ROYAL}`;
 
 /**
+ * match for "decision of the flemish government"
+ */
+const DECISION = 'besluit';
+
+/**
  * The type of citation that we need to search for
  */
-const TYPE = `${DECREE}|${MEMO}|${TREATY}|${CONSTITUTION_CHANGE}|${COLLAB}|${BOOK}|${PROTOCOL}|${FLEMGOV}|${COORD}|${SPECIAL}|${LAW}|${ROYAL}|${MINISTERIAL}|${ENUM_ROYAL}`;
+const TYPE = `${DECREE}|${MEMO}|${TREATY}|${CONSTITUTION_CHANGE}|${COLLAB}|${BOOK}|${PROTOCOL}|${FLEMGOV}|${COORD}|${SPECIAL}|${LAW}|${ROYAL}|${MINISTERIAL}|${ENUM_ROYAL}|${DECISION}`;
 /**
  * The monster regex that makes the citation plugin trigger to show `CitationCard`.
  * In restructuring, I've made sure that I didn't abstract away any of the capturing groups,
--- a/addon/plugins/citation-plugin/utils/process-match.ts
+++ b/addon/plugins/citation-plugin/utils/process-match.ts
@@ -1,5 +1,5 @@
 import { isBlank } from '@ember/utils';
-import { LEGISLATION_TYPES } from './types';
+import { isLegislationType, LEGISLATION_TYPES } from './types';
 import { unwrap } from '@lblod/ember-rdfa-editor-lblod-plugins/utils/option';
 
 const STOP_WORDS = ['het', 'de', 'van', 'tot', 'dat'];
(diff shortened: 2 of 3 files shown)


addon/components/citation-plugin/citations/decision-preview.ts
```diff
@@ -0,0 +1,19 @@
...
+} from '@lblod/ember-rdfa-editor-lblod-plugins/plugins/citation-plugin/utils/legal-documents';
+
+interface Args {
+  decision: LegalDocument;
+}
+
+export default class DecisionPreviewComponent extends Component<Args> {
+  get isDecision() {
+    return isBesluitLegalDocument(this.args.decision);
+  }
```
x-m-el, line 14: example of where the term "decision" is used in two different contexts. It is a `DecisionPreview` component, so you'd expect `isDecision` would always be true. But they both refer to a different type of "decision".

addon/components/citation-plugin/citations/decision-preview.ts
```diff
@@ -0,0 +1,19 @@
...
+
+  get hasPublicationLink() {
+    return this.isDecision && this.args.decision.meta?.publicationLink;
+  }
```
x-m-el, line 18: this is not used anywhere

addon/plugins/citation-plugin/utils/legal-documents.ts
```diff
@@ -0,0 +1,133 @@
...
+
+    const fetchConfig = {
+      words,
+      filter,
+      pageNumber,
+      pageSize,
+      config: {
+        endpoint: shouldQueryPublicDecisions
+          ? (config.decisionsEndpoint as string)
+          : config.endpoint,
```
x-m-el, line 120: if decisionsEndpoint is not provided, it will search for decisions on the other endpoint, which in our case will never return any. In a way this could make sense: a vendor might have a backend that provides everything (including decisions) in one endpoint.

However, I don't think this is a good idea:
- if this is the case with vendors, they can still define the same url for `decisionsEndpoint` and ` endpoint`
- I think it will always be separate, because all vendors will be using the same codex flemish codex url. The code for citation is too specific for this codex to really be used with anything else.
- vendors wanting to just use the codex (and no self-published decisions) will always see "besluit" in the dropdown, which will never return anything.

So I'd change the logic so that if `config.decisionsEndpoint` is not defined, the dropdown does not show `Besluit` as an option to choose from.

x-m-el (changes requested):
- it is a bit weird that the link for a published decision doesn't actually work, while for other citations it does. But I guess it is what it is :shrug: 
- The citation plugin uses "decision" as a general term for what it citates, but with your current implementation, you give "decision" a more specific meaning too and have used `LegalDocument` most of the time instead.
E.g. in decision-preview.hbs there is `this.isDecision`.
Not sure what the best way would be to make this code less chaotic. As far as I can see you're using `legalDocument` as everything cite-able?
so either make all `general decision` usages use `legalDocument` instead (also the file names), or use `besluit` for a "specific decision", instead of also the term "decision".
Now it is a bit of both together.
- should update the readme toe reflect the changes and the extra parameter in the config.



- I could not get `citation-card` to show up to test it. Seems like it just shows up in your video. Couldn't figure out how to show it through the code either, because it uses some special `Decorations` logic and honestly I can't follow what's happening there :p 

- didn't try to *break* this yet. Mostly did code reviewing first.

x-m-el (changes requested):
note: I pushed some updated documentation, so make sure you pull first.

- In a way, it would make sense to also make this work with the use of keywords as defined in `addon/plugins/citation-plugin/utils/process-match.ts`. However, this might be hard to update (although the code does look okay). So I'm not sure if it is worth the effort. You could try and see if it isn't too hard though.
A keyword could just be "besluit" I think.

- The button for "not public" is different than if it is active. Is there a specific design reason you don't use `@icon='link-external'` for the "full content not public" button, or just an oversight?

- Okay for me to keep the restructuring of code in separate PR. Would indeed make it hard to follow everything if files get renamed. Github doesn't always handle that perfectly.

Check out the "keyword typing search" thing mentioned above. For the rest I couldn't get it to break and looks good :+1:

PR #281, opened 2023-09-04
GN-4483: Use `renderInPlace` for all power select calls

dkozickis:
### Overview

Use `renderInPlace` attribute for all `PowerSelect` calls.

As we are not in control of the wrapper where the embedded editor is rendered, it is safer to use `renderInPlace` for `PowerSelect`, there is some cons (no automatic position detection) as described here https://ember-power-select.com/docs/troubleshooting (see `renderInPlace` part) but I think the pros (select is rendered near where it should be, below the select target) are outweighing the cons. 

https://binnenland.atlassian.net/browse/GN-4483

### Setup

1. Checkout
2. `npm run start`

### How to test/reproduce

.... WIP ....

### Checks PR readiness
- [ ] changelog
- [ ] npm lint
- [ ] no new deprecations

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+## Fixed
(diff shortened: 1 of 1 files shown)


x-m-el:
Not sure if you read the report (as it was in Dutch), but it mentions adding `demo.css` (found in public folder). Not clear if it fully fixed the bug or they just tried it out, so note this might be full red-herring :) 
Could see if that css has a simpler fix or something :shrug:

PR #286, opened 2023-09-06
GN-4266: Filter by government name

dkozickis:
### Overview

Builds on top of https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/275 to add filter by government name. 

https://binnenland.atlassian.net/browse/GN-4266

### Setup

1. Checkout
2. `npm run start`

### How to test/reproduce

See the video

https://github.com/lblod/ember-rdfa-editor-lblod-plugins/assets/769698/5e896602-71aa-441f-80ce-09f60475569e

### Challenges/uncertainties

Instead of providing a dropdown to select a government I decided it's better to allow user to enter the name, this absolves out from figuring out how to fetch the correct names. It also allows user to provide the default filter easier, as there is no need to map from a government ID to a government name.

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/translations/nl-BE.yaml
+++ b/translations/nl-BE.yaml
@@ -70,6 +70,7 @@ citaten-plugin:
     title: Citeeropschrift invoegen
   search:
     term: Zoekterm
+    government-search: Filter op overheid
     advanced: Uitgebreid zoeken
     type: Type van document
     placeholder: Typ om te zoeken
--- a/translations/nl-BE.yaml
+++ b/translations/nl-BE.yaml
@@ -70,7 +70,7 @@ citaten-plugin:
     title: Citeeropschrift invoegen
   search:
     term: Zoekterm
-    government-search: Filter op overheid
+    government-search: Filter op bestuursorgaan
     advanced: Uitgebreid zoeken
     type: Type van document
     placeholder: Typ om te zoeken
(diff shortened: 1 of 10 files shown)


translations/nl-BE.yaml
```diff
@@ -70,6 +70,7 @@ citaten-plugin:
     title: Citeeropschrift invoegen
   search:
     term: Zoekterm
+    government-search: Filter op overheid
```
abeforgit, line 73:
```suggestion
    government-search: Filter op bestuursorgaan
```

abeforgit (changes requested):
If I filter on Aalst, it doesn't find anything, even though there seem to be uittreksels published on QA for aalst?
image

abeforgit:
yeah so im afraid those also contain decisions
to an extract is a treatment (behandeling-van-agendapunt) but those (usually) contain decisions. 
But from a quick look at your query I think you're just overcomplicating things? no need to involve the decisionlist at all, just look for things with type `besluit:Besluit`, that's really all we care about.
Some decisions will have full content, (the extract is published), some won't (they are only published as being part of a decisionlist) 
but the citation plugin should show either.

PR #289, opened 2023-09-07
Move to changesets for changelog management

elpoelma:
### Overview
Adds [changesets](https://github.com/changesets/changesets) as a tool for changelog management.
Refer to their docs for details, but essentially, instead of manually adding an entry
to the changelog file, you can now do `npx changeset` and it will create a changeset file.

Because it's a separate file, there will be no more merge conflicts relating to changelogs, and
having a cli tool is nice.


x-m-el (approved):
as far as I can see looks correct.

What is missing (and I think also on ember-rdfa-editor) is potentially an update to the readme/contribution guide how to use this? But as that is still the default template I guess it's not too useful.
The .changeset folder does include a link to the documentation.

PR #290, opened 2023-09-07
GN-4479: make address node work properly inside a location variable

elpoelma:
### Overview
This PR adds support for inserting/editing address variables inside location variables. 
Specifically, it includes the following changes:
- The address node-view has been added to the embedded-editor of location variables
- Several css changes to ensure it remains clear to the user which variable is selected

[GN-4479](https://binnenland.atlassian.net/browse/GN-4479?atlOrigin=eyJpIjoiYWVmMTZlNzhhODdjNDZhZTk4YmNkYzJjY2ZiZDhlZWUiLCJwIjoiaiJ9)

### How to test/reproduce
- Open the regulatory statements dummy page
- Insert a location variable and enter it's embedded editor
- Insert an address variable inside the location variable
- Fill in the address variable
- Reload to check if all variables are correctly persisted

### Challenges/uncertainties
- On styling
   * When selecting an address variable inside a location variable and then leaving the location variable, the address variable still seems selected.
   * Unsure if it is clear enough to users which variable is selected (including the distinction between atomic and non-atomic variables)
- When a location variable only contains and address variable, the cursors don't work as expected. Including the firefox and chrome cursor plugins inside the embedded-editor gives weird results.

### Checks PR readiness
- [ ] UI
- [x] changelog
- [x] npm lint
- [x] no new deprecations


elpoelma:
> Clicking in the nested node and then clicking outside into the document does not remove the "selected" outline, not functionality breaking and not a blocker, but I think UI wise this can be improved
> CleanShot.2023-09-13.at.14.33.39.mp4

Yes I think so too, this was already an issue before with other nodes inside an embedded editor. I'm not sure what the best way is to tackle this issue (through css or js).

abeforgit:
the issue is that the inner node doesn't lose its "selected" state (I assume because its an inner editor that keeps its own selection)
this means it never loses the Prosemirror-selectedNode class

I guess when you click outside an embedded editor, we clear that state, but only for the "first one"?

elpoelma:
> the issue is that the inner node doesn't lose its "selected" state (I assume because its an inner editor that keeps its own selection) this means it never loses the Prosemirror-selectedNode class
> 
> I guess when you click outside an embedded editor, we clear that state, but only for the "first one"?

I don't think we ever clear that state tbh

elpoelma:
On the main view of the editor, we could do something like this:
```js
this.mainView = new SayView(target, {
      ...
      handleDOMEvents: {
        focus: () => {
          if (this.mainView !== this.activeView) {
            const tr = this.activeView.state.tr;
            tr.setSelection(TextSelection.create(this.activeView.state.doc, 0));
            this.activeView.dispatch(tr);
            this.setActiveView(this.mainView);
          }
        },
      },
      ...
    });
    ```
When we are refocusing the main view, clear the selection (text-selection at the beginning) of the embedded view.

abeforgit:
> On the main view of the editor, we could do something like this:
> 
> this.mainView = new SayView(target, {
>       ...
>       handleDOMEvents: {
>         focus: () => {
>           if (this.mainView !== this.activeView) {
>             const tr = this.activeView.state.tr;
>             tr.setSelection(TextSelection.create(this.activeView.state.doc, 0));
>             this.activeView.dispatch(tr);
>             this.setActiveView(this.mainView);
>           }
>         },
>       },
>       ...
>     });
> 
> When we are refocusing the main view, clear the selection (text-selection at the beginning) of the embedded view.

sounds sensible, but you can merge this pr without for now

abeforgit:
issue tracked here https://binnenland.atlassian.net/browse/GN-4531

PR #296, opened 2023-09-11
bump ember deps and add them to peerDeps

abeforgit:
### Overview

ember source and data are peerdeps, so we should add them as such. As a bonus, bumps ember and ED
for the dev setup to 4.12 cause why not.

EDIT about 4 hours later:
this turned into quite the rabbithole. Essentially I wanted to test the new setup in GN, since I
wasn't convinced the standard template plugin was being hit in the dummy app
So naturally I `npm link`ed the plugins there.

To my surprise I got a runtime error from ember-resources. It seemed to think that it now lived in
a 4.12 app, even though GN is 4.8. So after some digging I figured this could be due to
the awkwardness of `npm link`.

I found [yalc](https://www.npmjs.com/package/yalc) which seemed promising. It _actually_ publishes
your package to a "local registry" and then allows to "really" install in in your host.
This ran into another issue, something about typescript `declare`d class properties not being
transpiled.

So I turned to discord and invoked the nullvox card. After some discussion I confirmed `npm link`
was likely the culprit of the first issue. Still, I had no way to actually link the two.
So I kept digging, and by chance stumbled on
[this issue](https://github.com/typed-ember/ember-cli-typescript/issues/1568).
Turns out ember-cli-typescript seems to be on the way out. Simplifying the setup couldn't hurt
(never really liked ember-cli-typescript, does way too much magic). To my surprise that actually
worked! I could now use either `yalc add` or `yalc link`, both working as intended.

After some debate I found out I can't quite get rid of e-cli-ts, cause the precompile and clean
commands are still necessary, and are quite involved pieces of scripting. This was annoying,
cause adding back the dep, even with the extra babel config, broke yalc again.
With some more struggling I finally got the idea to keep it, but as a devdep. This makes it so
the cli commands are available for yalc but the dep won't be included in the package, so it can't
do whatever it was doing to the build pipeline that was breaking things. So win-win.

### How to test/reproduce

So as mentioned, you kinda need yalc.
It's very easy to use and I think you'll see a lot more of it in the near future, it could
potentially do wonders for dependent PR misery.

run `yalc publish` in this repo
run `yalc link @lblod/ember-rdfa-editor-lblod-plugins` in GN

verify the templates work and nothing else breaks.

### Challenges/uncertainties

uncertainties abound, I have no clue what is going on with this mad world of package managers
and peerdeps

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/sharp-cobras-act.md
+++ b/.changeset/sharp-cobras-act.md
@@ -1,0 +1,5 @@
+---
+'@lblod/ember-rdfa-editor-lblod-plugins': minor
+---
+
+bump ember and ember data to 4.12 for development, and add correct peerdep specification for consumers
--- a/README.md
+++ b/README.md
@@ -5,7 +5,9 @@ related to the LBLOD Project.
 
 ## Compatibility
 
-* Ember.js v3.28 or above
+* Ember.js and ember data v3.28 or 4.x
+note: we smoke-test for 3.28 but develop on the latest 4.x minor. The 5.x range is currently untested and not officially supported, but we accept issues and PRs to do so.
+
 * Embroider or ember-auto-import v2
 * Node 18 or above
 
--- a/index.js
+++ b/index.js
@@ -6,6 +6,9 @@ module.exports = {
     return process.env.EMBER_ENV === 'development';
   },
   options: {
+    'ember-cli-babel': {
+      enableTypeScriptTransform: true,
+    },
     babel: {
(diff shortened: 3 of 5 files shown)


elpoelma (approved):
Seems to work well on first sight. Tested with latest master of GN, ember-data does not seem to have any issues.

PR #298, opened 2023-09-15
Address Variable: add specific variable insert component that includes label logic

x-m-el:
### Overview
Address variable had an insert component for inserting it via the general "insert menu". However, this did not allow adding a label and the button is different from that of other variables.
This adds a specific insert-variable component meant to be used in the variable dropdown component.
- has the same button as other variable inserts
- also allows inputting a label

found while updating embeddable. Not connected ticket.

### How to test/reproduce
The address variable is added to the list of variables in the dropdown.
Add it, see that the button is the same as the others, see that setting a label works.
also look at the other insert button (under the insert menu): see that this still works.

### Challenges/uncertainties
I don't think the default label is translated at this point. This is probably what is getting fixed in #297 
As usual, depending on which one gets finished/reviewed first, this address variable also needs the same translating logic.

general logic is extracted in a `utils.ts` file. Awful name, but it is very close to the usecase (in same folder), so should be fine.

### DIFF
--- a/.changeset/violet-hats-sneeze.md
+++ b/.changeset/violet-hats-sneeze.md
@@ -1,0 +1,8 @@
+---
+'@lblod/ember-rdfa-editor-lblod-plugins': minor
+---
+
+Added a second address insert `variable-plugin/address/insert-variable`. This replaces the `insert` from before when used inside the `insert-variable-card` dropdown.
+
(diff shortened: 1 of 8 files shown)


elpoelma (approved):
Looks good!

PR #302, opened 2023-09-18
GN-4517: Reduce possible CSS conflicts

dkozickis:
### Overview

* Mostly add `say-` prefix for CSS classes that sound like they are too generic.
* Remove unused `span.date` class

https://binnenland.atlassian.net/browse/GN-4517

### Setup

1. Checkout

### How to test/reproduce

Make sure classes are still applying where they should be
### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/tricky-needles-shake.md
+++ b/.changeset/tricky-needles-shake.md
@@ -1,0 +1,5 @@
+---
+'@lblod/ember-rdfa-editor-lblod-plugins': patch
+---
+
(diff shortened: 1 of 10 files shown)


x-m-el:
I think we have the code that they had this css overlap problem with. Were you able to check out that code to see if you could find what the overlap might be?
It was posted somewhere in rocketchat but I'm not sure anymore where :sweat_smile:

PR #307, opened 2023-09-25
GN-4502: allow municipality and street field of address plugin to be cleared

elpoelma:
### Overview
This PR ensures that users are able to clear the municipality and street inputs when editing an address variable. This is implemented by passing the `allowClear` argument to the `power-select` components.

[GN-4502](https://binnenland.atlassian.net/browse/GN-4502?atlOrigin=eyJpIjoiMzRhYzI1NjI3Nzc3NDQ3M2I3ZDkyMjVkZTcyZDIwYjIiLCJwIjoiaiJ9)

### How to test/reproduce
- Insert an address variable and select it
- Notice that when the street/municipality fields are filled out, you can clear their value
- When clearing the street input, the housenumber and busnumber fields should also be cleared
- When clearing the municipality input, all other fields should also be cleared

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/nine-otters-behave.md
+++ b/.changeset/nine-otters-behave.md
@@ -1,0 +1,5 @@
+---
+'@lblod/ember-rdfa-editor-lblod-plugins': patch
+---
+
+Allow municipality and street fields of address-variable to be cleared when editing
--- a/addon/components/variable-plugin/address/edit.hbs
+++ b/addon/components/variable-plugin/address/edit.hbs
@@ -11,6 +11,7 @@
           @searchMessage={{t "editor-plugins.address.edit.municipality.search-message"}}
           @noMatchesMessage={{t "editor-plugins.address.edit.municipality.no-results"}}
           @placeholder={{t "editor-plugins.address.edit.municipality.placeholder"}}
+          @allowClear={{true}}
           @renderInPlace={{true}}
           @searchEnabled={{true}}
           @search={{perform this.searchMunicipality}}
@@ -27,6 +28,7 @@
           @searchMessage={{t "editor-plugins.address.edit.street.search-message"}}
           @noMatchesMessage={{t "editor-plugins.address.edit.street.no-results"}}
           @placeholder={{t "editor-plugins.address.edit.street.placeholder"}}
+          @allowClear={{true}}
           @renderInPlace={{true}}
           @searchEnabled={{true}}
           @search={{perform this.searchStreet}}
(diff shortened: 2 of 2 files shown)


x-m-el (approved):
small bug
1. open the street select
2. now click on the "clear city" cross
3. select of street will still be highlighted (blue border), but disabled.
Not sure if the problem comes from powerselect and if it is fixable. But maybe you can hook into the clearing and deselect it or change the css so it doesn't hightlight something disabled? 
Problem does not occur for the text inputs.

No other issues I could find :)

elpoelma:
> small bug
> 
>     1. open the street select
> 
>     2. now click on the "clear city" cross
> 
>     3. select of street will still be highlighted (blue border), but disabled.
>        Not sure if the problem comes from powerselect and if it is fixable. But maybe you can hook into the clearing and deselect it or change the css so it doesn't hightlight something disabled?
>        Problem does not occur for the text inputs.
> 
> 
> No other issues I could find :)

I think we can classify this as a power-select bug, as the `ember-power-select-trigger--active` class is not removed when the select-element is disabled. We could indeed fix it with css, I don't know if it is fixed in the latest release of `power-select` (7.x.x)

elpoelma:
> > small bug
> > ```
> > 1. open the street select
> > 
> > 2. now click on the "clear city" cross
> > 
> > 3. select of street will still be highlighted (blue border), but disabled.
> >    Not sure if the problem comes from powerselect and if it is fixable. But maybe you can hook into the clearing and deselect it or change the css so it doesn't hightlight something disabled?
> >    Problem does not occur for the text inputs.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > No other issues I could find :)
> 
> I think we can classify this as a power-select bug, as the `ember-power-select-trigger--active` class is not removed when the select-element is disabled. We could indeed fix it with css, I don't know if it is fixed in the latest release of `power-select` (7.x.x)

It does not seem to be fixed in the latest 7.x release. I could open an issue on appuniversum, to not apply the blue border style when the input is disabled.

x-m-el:
Issue on AU makes sense (and more chance it'll actually get fixed).
It's not a breaking thing so I think that's good enough. At least it'll be fixed sometime in the future then :)

Could add a comment to the code linking to the issue, so it is obvious the bug is known and isn't researched uselessly in the future.

elpoelma:
> Issue on AU makes sense (and more chance it'll actually get fixed). It's not a breaking thing so I think that's good enough. At least it'll be fixed sometime in the future then :)
> 
> Could add a comment to the code linking to the issue, so it is obvious the bug is known and isn't researched uselessly in the future.

I submitted an issue: https://github.com/appuniversum/ember-appuniversum/issues/439. I did notice it is not an issue on chromium-based browsers. So the fact that the class is not removed/the input-component is not deselected could be a firefox bug.

PR #313, opened 2023-09-29
GN-4525 Remove ember-data dependency

piemonkey:
### Overview
Remove usage and types from ember-data. Standard Template service now requires a list of templates to be passed in (breaking change). The types of some objects used from frontend-gelinkt-notuleren are now less specific to allow for the functionality that is currently used but without the ember-data dependency.

It feels a bit weird that we're using types from a consumer project, but I guess these are around for historical reasons.

https://binnenland.atlassian.net/browse/GN-4525
[frontend PR](https://github.com/lblod/frontend-gelinkt-notuleren/pull/583)

### Setup
Needs the latest `frontend-gelinkt-notuleren` as that passes in the templates correctly.

### How to test/reproduce
Test that editing agenda points and inserting templates works as expected.

### Challenges/uncertainties
N/A

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations


addon/models/template.ts
abeforgit, line 1: this file should move to the types dir, the `models` dir has special meaning in ember so best not to use it

types/lblod/frontend-gelinkt-notuleren/models/account.d.ts
abeforgit, line 1: gonna be honest, I had no idea this dir even existed, I think these files should all be deleted, along with the session service altogether
I can see the besluittype plugin still uses them, so let's leave them in in this pr and make another one to normalize that plugin

piemonkey: When you say normalise that plugin, I assume you mean to pass in the `classificatie.uri` which looks to be the only thing it's getting from the session service? (along with anything else it turns out it also needs)

PR #322, opened 2023-10-10
Make plugins follow document language where applicable (show as words & placeholders)

x-m-el:
### Overview
There was a ticket for already partially using document language for plugins: https://binnenland.atlassian.net/browse/GN-4450

This fixes all other instances where translations still had to happen.
The clue is that anything that shows up in the document (part of the "reading" experience") should be in the language of the document.

Number variable: interestingly, a very recent n2words added support for a general language as locale. So no special logic needed for that
Placeholders: placeholders are part of the document and are translated with doc language.

https://binnenland.atlassian.net/browse/GN-4424

### How to test/reproduce
- Check if number as words translates based on the doc language
- Check if placeholders follow this language (e.g. article structures, decision structures).
- This PR is a bit rushed, so I might have missed some plugins that should still change to document language. Part of testing is checking if this PR fixes all instances so the linked issue can indeed be closed.

### Challenges/uncertainties
- The name for artikel, hoofdstuk, sectie etc are not translated and shouldn't. These are used specifically in regexes and are linked to official names for legal documents. So changing these would need more insight if that is correct and changes to the regexes.
- I tried to make this as non-breaking as possible
  - The code uses a new plugin that might not be configured. If this is the case, old logic is used (translating with browser locale).
  - This does meant `intl` still has to be passed. However, notice that this is anyway needed to give a default translation and removing that parameter means changing the `StructureSpec` api (which would be a breaking change)
  - `StructureSpec` gets an extra parameter, but old parameters are not changed, which keeps backwards compatibility with article-structure's old logic. In theory a user of this plugin could have specified their own `StructureSpec`s.

### DIFF
--- a/.changeset/khaki-schools-clap.md
+++ b/.changeset/khaki-schools-clap.md
@@ -1,0 +1,5 @@
+---
+'@lblod/ember-rdfa-editor-lblod-plugins': patch
+---
+
+Using "show as words" for a number variable will convert the number to a string in the language in the document, instead of always showing Dutch.
--- a/.changeset/smart-fireants-sniff.md
+++ b/.changeset/smart-fireants-sniff.md
@@ -1,0 +1,11 @@
+---
+'@lblod/ember-rdfa-editor-lblod-plugins': minor
+---
+
+For article-structure plugin
+
+- The `StructureSpec`'s `constructor` now also contains the optional argument `state` (an EditorState)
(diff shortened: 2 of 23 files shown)


elpoelma (approved):
Code looks sensible and this PR seems to resolve most/all? of the remaining missing translations/translation issues.

I think it is indeed sensible to not translate the legal terms (like 'artikel' etc.)

PR #334, opened 2023-10-20
Extend support for `ember-concurrency` version to 3.1 and higher

elpoelma:
### Overview
This PR extends version range of the required `ember-concurrency` peer dependency to include the 3.x range on top of the 2.x range.

### Challenges/uncertainties
- Notice that aside from being specified as a peerdep, `ember-concurrency` is also installed as a hard-dep (instead of a dev-dep). The `ember-resources` throws an error when having it installed as a devdep for some reason...

### Checks PR readiness
- [ ] changelog
- [ ] npm lint
- [ ] no new deprecations


package.json
```diff
@@ -146,7 +146,7 @@
     "@appuniversum/ember-appuniversum": "^2.4.2",
     "@glint/template": "^1.1.0",
     "@lblod/ember-rdfa-editor": "^6.0.0",
-    "ember-concurrency": "^2.3.7",
+    "ember-concurrency": "2.x || ^3.1.0",
```
abeforgit, line 149: any reason you didn't add the `^` here?

PR #358, opened 2023-11-28
Update article structure plugin to support rdfa editing

piemonkey:
### Overview
Correctly support the insertion of any structures in the article structures plugin.

All comments should have been addressed, with the exception of pressing enter in the titles, which still creates a whole new structure with a new title. There are some design issues, but since we'll need to overhaul the design for editable nodes, I haven't done any tweaking.

This is made towards #357 with the intent that it collects all of the related changes rather than them being merged straight to master.
[PR on editor repo](https://github.com/lblod/ember-rdfa-editor/pull/1062)

### Setup
The 'editable nodes' route should now have the article structure plugin enabled.

### How to test/reproduce
Inserting structure elements should function as expected.

### Checks PR readiness
- [ ] changelog
- [ ] npm lint
- [ ] no new deprecations

### DIFF
--- a/addon/plugins/article-structure-plugin/structures/article.ts
+++ b/addon/plugins/article-structure-plugin/structures/article.ts
@@ -39,7 +39,7 @@ export const articleSpec: StructureSpec = {
     const titleRdfaId = uuid();
     const headingRdfaId = uuid();
     const numberRdfaId = uuid();
-    const bodyUuid = uuid();
+    const bodyRdfaId = uuid();
     const resource = `http://data.lblod.info/articles/${articleUuid}`;
     const titleText = translationWithDocLang(
       PLACEHOLDERS.title,
@@ -68,13 +68,13 @@ export const articleSpec: StructureSpec = {
         {
           type: 'external',
           predicate: SAY('body').prefixed,
-          object: { type: 'literal', rdfaId: bodyUuid },
+          object: { type: 'literal', rdfaId: bodyRdfaId },
         },
       ],
       backlinks: [],
     };
     const bodyAttrs: RdfaAttrs = {
-      __rdfaId: bodyUuid,
+      __rdfaId: bodyRdfaId,
       rdfaNodeType: 'literal',
       backlinks: [
         {
@@ -110,15 +110,13 @@ export const articleSpec: StructureSpec = {
           ),
       ),
     ]);
-    const selectionConfig: {
-      relativePos: number;
-      type: 'text' | 'node';
-    } = content
-      ? { relativePos: 5, type: 'text' }
-      : { relativePos: 6, type: 'node' };
+
     return {
       node,
-      selectionConfig,
+      selectionConfig: {
+        type: content ? 'text' : 'node',
+        rdfaId: bodyRdfaId,
+      },
       newResource: resource,
     };
   },
--- a/addon/plugins/article-structure-plugin/structures/article.ts
+++ b/addon/plugins/article-structure-plugin/structures/article.ts
@@ -37,6 +37,7 @@ export const articleSpec: StructureSpec = {
     const translationWithDocLang = getTranslationFunction(state);
     const articleUuid = uuid();
     const titleRdfaId = uuid();
+    const headingRdfaId = uuid();
     const bodyUuid = uuid();
     const resource = `http://data.lblod.info/articles/${articleUuid}`;
     const titleText = translationWithDocLang(
@@ -54,9 +55,9 @@ export const articleSpec: StructureSpec = {
           object: numberConverted,
         },
         {
-          type: 'attribute',
+          type: 'external',
           predicate: SAY('heading').prefixed,
-          object: `Artikel ${numberConverted}: ${titleText}`,
+          object: { type: 'literal', rdfaId: headingRdfaId },
         },
         {
           type: 'external',
@@ -87,6 +88,7 @@ export const articleSpec: StructureSpec = {
         backlinkResource: resource,
         titleRdfaId,
         titleText,
+        headingRdfaId,
         number: numberConverted,
         headerType: 'article_header',
       }),
--- a/addon/plugins/article-structure-plugin/structures/article.ts
+++ b/addon/plugins/article-structure-plugin/structures/article.ts
@@ -38,6 +38,7 @@ export const articleSpec: StructureSpec = {
     const articleUuid = uuid();
     const titleRdfaId = uuid();
     const headingRdfaId = uuid();
+    const numberRdfaId = uuid();
     const bodyUuid = uuid();
     const resource = `http://data.lblod.info/articles/${articleUuid}`;
     const titleText = translationWithDocLang(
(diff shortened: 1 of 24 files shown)


addon/plugins/article-structure-plugin/structures/article.ts
```diff
@@ -34,37 +35,77 @@ export const articleSpec: StructureSpec = {
...
+      properties: [
+        {
+          type: 'attribute',
+          predicate: ELI('number').prefixed,
+          object: numberConverted,
+        },
+        {
+          type: 'attribute',
+          predicate: SAY('heading').prefixed,
+          object: `Artikel ${numberConverted}: ${titleText}`,
```
elpoelma, line 59: What is the reason that these are still added as attribute-properties? As their content would also be included in the title-node right?

addon/plugins/article-structure-plugin/structures/chapter.ts
```diff
@@ -28,37 +34,76 @@ export const chapterSpec: StructureSpec = {
...
+        {
+          type: 'attribute',
+          predicate: ELI('number').prefixed,
+          object: numberConverted,
+        },
+        {
+          type: 'attribute',
+          predicate: SAY('heading').prefixed,
+          object: `${numberConverted}. ${titleText}`,
+        },
```
elpoelma, line 59: Same remark as above.

addon/plugins/article-structure-plugin/structures/structure-header-title.ts
```diff
@@ -0,0 +1,38 @@
+import { NodeSpec } from 'prosemirror-model';
+import { getRdfaAttrs, rdfaAttrSpec } from '@lblod/ember-rdfa-editor';
+import { renderRdfaAware } from '@lblod/ember-rdfa-editor/core/schema';
+import { EXT } from '@lblod/ember-rdfa-editor-lblod-plugins/utils/constants';
+
+export const structure_header_title: NodeSpec = {
+  attrs: {
+    ...rdfaAttrSpec,
+    __tag: { default: 'span' },
```
elpoelma, line 9: I don't think this attribute is necessary anylonger.

addon/plugins/article-structure-plugin/utils/structure.ts
```diff
@@ -38,26 +45,24 @@ export function constructStructureNodeSpec(config: {
...
         },
-        0,
-      ];
+        content: 0,
+      });
     },
     parseDOM: [
       {
         tag: 'div',
+        priority: 55,
```
elpoelma, line 61: Could you add a comment as to why this priority is added/needed?

addon/plugins/article-structure-plugin/utils/structure.ts
```diff
@@ -38,26 +45,24 @@ export function constructStructureNodeSpec(config: {
...
         getAttrs(element: HTMLElement) {
-          if (
-            hasRDFaAttribute(element, 'property', SAY('hasPart')) &&
-            hasRDFaAttribute(element, 'typeof', type) &&
-            element.getAttribute('resource')
-          ) {
-            return { resource: element.getAttribute('resource') };
+          const resource = element.getAttribute('resource');
+          if (hasRDFaAttribute(element, 'typeof', type) && resource) {
```
elpoelma, line 64: I suppose we could do this check on the properties attribute of the incoming html-element now? 
e.g.:
```js
attrs.properties.some((prop) => prop.predicate === 'typeof' && prop.object.resource === type)
```

addon/plugins/article-structure-plugin/utils/structure.ts
```diff
@@ -99,6 +109,89 @@ export function constructStructureBodyNodeSpec(config: {
...
+            : {}),
+        },
+        ...numberContentDOM(node.attrs.number),
+        ['span', {}, 0],
+      ];
+    },
+    parseDOM: [
+      {
+        tag: 'h1,h2,h3,h4,h5,h6,span',
+        priority: 60,
```
elpoelma, line 179: Could you add a comment on the reason of this `priority`?

addon/plugins/article-structure-plugin/utils/structure.ts
```diff
@@ -99,6 +109,89 @@ export function constructStructureBodyNodeSpec(config: {
...
+        getAttrs(element: HTMLElement) {
+          const level = TAG_TO_LEVEL.get(element.tagName.toLowerCase()) ?? 6;
+          const property = element.getAttribute('property');
```
elpoelma, line 182: We could do this check with the `properties` and `backlinks` attributes now I think.

addon/plugins/standard-template-plugin/utils/nodes.ts
elpoelma, line 1: I think we can also update the `parseDOM` of these nodes to leverage the `properties` and `backlinks`. The `toDOM` should also be updated to use the `renderRdfaAware` function. I'm also not sure if it will be necessary to keep all these seperate node-specs.

elpoelma (commented):
The different article structures seem to be well integrated with the new rdfa-system already :) Very nice work!! I have added some comments with questions/remarks.

Some additional questions/remarks:
- There still seem to be some issues with selections in and around structures, some of which also are related to the gap-cursor. We should look into improving that behaviour.
- (not new) Pressing `enter` in a title splits the node in two, ideally `enter` should do nothing.
- The different article structures have each two attribute-properties (heading and number), for which I'm not sure if these are (still) needed.
- (design-related, not specific to this PR): we should probably make it clearer which parts of the structures are editable, and which are not (e.g. with different backgrounds, different cursors). Probably needs design input.
- We could look into leveraging the `properties` and `backlinks` attributes in the parseDOM-rules in order to parse the different structures.

addon/plugins/article-structure-plugin/utils/structure.ts
piemonkey: I'm not sure what you mean. Do you mean to use `getRdfaAttrs()` to process the element and use the `properties` member of that attrs object?

addon/plugins/article-structure-plugin/utils/structure.ts
piemonkey: Added

addon/plugins/article-structure-plugin/utils/structure.ts
piemonkey: I removed this one actually as it wasn't necessary...

addon/plugins/article-structure-plugin/structures/article.ts
piemonkey: I've pushed a version with heading split out, I'll put one in with number as an external tomorrow.

addon/plugins/article-structure-plugin/utils/structure.ts
abeforgit: note the schema you define for a doc is also priority-ordered, so if we put the structure nodes before the headers, it should also parse correctly with default prio
(gives more flexibility to the host app)

PR #365, opened 2023-12-22
Make variables rdfa-aware

abeforgit:
### Overview

This PR integrates the variables with the new rdfa system of resource nodes and literal nodes.
In essence, all it does is
- add a new parserule for each variable, with a higher prio than the old one, which
uses the parsed rdfa attributes rather than rigidly checking for the expected structure.

- rewrite the toDom methods to use `renderRdfaAware`.
- update the insertion logic to make sure the correct rdfa attributes are inserted (like `properties`, `rdfaNodeType`, `subject`, and so on)
- update the editing logic to make sure the appropriate rdfa properties are updated

what this PR does NOT do is replace the pre-existing PM attributes that were used to manage the
variable state, it just adds the rdfa attributes alongside. If possible I would like to eventually
phase those out, it was just too much at once to do now.

All variables are resource nodes, with the `mappingUri` as their URI. Their values are
`ext:content`, which in somecases is a contentPredicate (like for the text variable), but
in others is a property with the `attribute` type. This reflects some variables having editable
content and others not.

The current structure does mean that it's currently not possible to link directly to a
variable's value from another resource. This is not a new limitation, but rather a consequence
of the chosen (or rather "accidentally grown") datamodel for variables.

Remains to be seen if this should be addressed at some point. (e.g., it's not really "correct" to
link a "startDate" predicate to the variable, cause the domain of a predicate like that is
likely a date literal)

TODOS:

- the address variable is not yet converted, all the others are
- backwards compatibility is not tested, although it should _theoretically_ "just" work.
- there are some issues regarding placeholder nodes. You can see this if you insert a date variable
and reload the page. The placeholder text is gone, but the variable still works fine.
- not quite sure about the way I manage the "content" of a codelist variable

needs this to work (`npm link` worked for me, didn't need `yalc`, ymmv):
[editor pr](https://github.com/lblod/ember-rdfa-editor/pull/1077)

### Checks PR readiness
- [ ] changelog
- [ ] npm lint
- [ ] no new deprecations


elpoelma:
### Progress/TODOS

#### Conversion of nodespecs
- [ ] text
- [ ] number
- [ ] date
- [ ] codelist
- [ ] location
- [ ] address

#### Backwards compatibility of nodespecs
- [ ] text
- [ ] number
- [ ] date
- [ ] codelist
- [ ] location
- [ ] address

#### Drop (obsolete) custom variable-attributes
- [ ] text
- [ ] number
- [ ] date
- [ ] codelist
- [ ] location
- [ ] address

elpoelma:
On the `location` variable: these contain an additional `plaatsbepaling` (https://data.vlaanderen.be/ns/mobiliteit#plaatsbepaling) predicate, not sure how to encode this yet.

abeforgit:
what do you mean exactly with "not sure how to encode"? what info do you need?

elpoelma:
As discussed, I reverted the changes to the `location` variable. We keep it as non rdfa-aware for now.
The only change is that the `https://data.vlaanderen.be/ns/mobiliteit#plaatsbepaling` is no longer included. We could still hardcode this in the `toDOM` of the location variable.

PR #369, opened 2024-01-15
GN-4650: Snippet selection placeholder in template

dkozickis:
### Overview

Use RDFA property to manage allowed snippets within a resource node. 

* If node has snippet list IDs stored in `properties` with predicate of `SNIPPET_LIST_RDFA_PREDICATE` (`say:snippetListIds`) then use those.
* If there are no snippet list IDs stored - scan upwards through resource nodes.

https://binnenland.atlassian.net/browse/GN-4650

### Setup

1. Checkout this branch
2. `npm link` or `yalc` the `feat/rdfa-links-ui` branch of `ember-rdfa-editor`

### How to test/reproduce

Have a look at the video here - https://share.cleanshot.com/dBwV9yNb

### Challenges/uncertainties

1. Not sure about predicate `say:snippetListIds`?

### Checks PR readiness
- [X] changelog
- [X] npm lint
- [X] no new deprecations

### DIFF
--- a/addon/components/snippet-plugin/snippet-insert-rdfa.ts
+++ b/addon/components/snippet-plugin/snippet-insert-rdfa.ts
@@ -57,7 +57,7 @@ export default class SnippetInsertRdfaComponent extends Component<Args> {
     while (parentNode) {
       const properties = getSnippetListIdsProperties(parentNode.node);
 
-      if (properties) {
+      if (properties && properties.length > 0) {
         return properties;
       }
 


abeforgit (changes requested):
nice work!
the one comment I have is to use multiple triples rather than one triple with a list as its value, its a bit more "linked data" that way

so a singular predicate, like `say:allowedSnippetList`, and rather than using the ID, use the full URI. And just add multiple instances of that triple when you add more lists

abeforgit (changes requested):
seems this broke the upward search? adding fragment lists to a resource node such as a structure now no longer allows them to be inserted in the body

addon/plugins/snippet-plugin/utils/rdfa-predicate.ts
```diff
@@ -0,0 +1,50 @@
+import { PNode } from '@lblod/ember-rdfa-editor';
+import { AttributeProperty } from '@lblod/ember-rdfa-editor/addon/core/rdfa-processor';
+import { getProperties } from '@lblod/ember-rdfa-editor/utils/_private/rdfa-utils';
+import { getSnippetIdFromUri } from '@lblod/ember-rdfa-editor-lblod-plugins/plugins/snippet-plugin';
+
+export const SNIPPET_LIST_RDFA_PREDICATE = 'say:snippetListIds';
```
elpoelma, line 6: As the triples are singular now, maybe it is better to use the `say:allowedSnippetList` as @abeforgit suggested?

abeforgit (changes requested):
now the button is always active even when there are no snippetlists configured

elpoelma:
Also found the following issue with the snippet-select modal:
[Screencast from 2024-01-29 16-41-16.webm](https://github.com/lblod/ember-rdfa-editor-lblod-plugins/assets/24712836/6f8f0594-870c-46c8-b0b2-e1a2c006afcf)

addon/components/snippet-plugin/snippet-list-select-rdfa.ts
```diff
@@ -0,0 +1,71 @@
...
+  get controller() {
+    return this.args.controller;
+  }
+
+  get node() {
+    return this.args.node;
+  }
+
+  get currentResource() {
+    return this.node.value.attrs.resource as string | undefined;
```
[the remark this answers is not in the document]
abeforgit: yep

addon/plugins/snippet-plugin/utils/rdfa-predicate.ts
```diff
@@ -0,0 +1,50 @@
...
+): AttributeProperty[] | undefined => {
+  const properties = getProperties(node);
+
+  if (!properties) {
+    return undefined;
+  }
+
+  return properties.filter(
+    (property): property is AttributeProperty =>
+      property.predicate === SNIPPET_LIST_RDFA_PREDICATE &&
```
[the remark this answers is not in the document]
elpoelma: Yep

addon/components/snippet-plugin/snippet-list/snippet-list-view.hbs
```diff
@@ -14,7 +14,7 @@
         {{#each @snippetLists as |snippetList|}}
           <tr>
             <td class="selectColumn">
-              <AuControlCheckbox
+              <AuCheckbox
                 @identifier={{snippetList.label}}
```
elpoelma, line 18: Is this argument still needed?

elpoelma (approved):
Seems to work well overall now :)

abeforgit:
> > now the button is always active even when there are no snippetlists configured
> 
> Admittedly, this was not discussed in the requirements, so I've went with the same approach as we use for non-RDFa snippet list attributes. I have now changed to hide the "Inserts snippets" button when there are no snippet lists assigned.

No, but you know we don't have the resources for fully worked out requirements, and all other buttons in that section disable themselves when they are not active.
I don't agree with hiding it either since it breaks the pattern, but you don't have to revert for that.
The reason the insert button was always active before is precisely because we could only specify lists for the entire document. If the button was also active when no lists were defined, then that was a UX bug that we missed during reviews.

addon/components/snippet-plugin/snippet-list/snippet-list-view.hbs
abeforgit: @dkozickis could you answer this please?

PR #376, opened 2024-01-18
GN-4652: Start number

dkozickis:
### Overview

* Allows to set a "start number" for a structure. 
* Introduce `numberDisplayStyle` and `startNumber` attributed for `_header` nodes.

https://binnenland.atlassian.net/browse/GN-4650

### Setup

1. Checkout
2. `npm run start`

### How to test/reproduce

Have a look at the video - https://share.cleanshot.com/d87pbryy

1. Try to reproduce what is happening in the video
2. Try checkout `master`, create a structure there, then checkout this branch and see if structure is still intact.

### Challenges/uncertainties

I have moved storing calculated `number` to the `content` attribute of "heading" element. So instead of storing "romanized" number, we now store actual integer (as string) and a `displayStyle` (on the root level). But, magically, this commit and current `master` and backwards compatible against each other, as we still store "converted" number (e.g. romanized) in the `textContent` of the "heading" element.  

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [X] no new deprecations


elpoelma:
> Not sure why the precompile fails :( Looks like we're checking `node_modules` TS files with our tsconfig, even though `skipLibCheck` is enabled 😅
> 
> P.S. OK it is the `ember ts:precompile` which uses own config for TS check. Wondering how did I break it in this PR.

The weird thing is that it doesn't fail on master :thinking:

addon/plugins/article-structure-plugin/utils/structure.ts
```diff
@@ -10,11 +10,19 @@ import {
...
   hasRDFaAttribute,
   Resource,
 } from '@lblod/ember-rdfa-editor-lblod-plugins/utils/namespace';
+import { unwrap } from '@lblod/ember-rdfa-editor-lblod-plugins/utils/option';
+import { StructureSpec } from '@lblod/ember-rdfa-editor-lblod-plugins/plugins/article-structure-plugin';
+import {
+  romanize,
+  romanToInt,
+} from '@lblod/ember-rdfa-editor-lblod-plugins/plugins/article-structure-plugin/utils/romanize';
+import { DOMOutputSpec } from 'prosemirror-model/src/to_dom';
```
elpoelma, line 25: The error seems to occur due to this line, which directly imports from a module in `prosemirror-model`. I would change this to an import from the `@lblod/ember-rdfa-editor` package.

elpoelma (commented):
Looks good technically, nice work :smile: 

Just wondering one thing:
At the moment, when the start-number of a structure is not explicitly set, it's current number is shown in the number-select. I'm wondering if we should also show to the user that the start-number is not explicitly set/undefined.

elpoelma (approved):
Looks good to me now!

PR #379, opened 2024-01-23
Rdfa/actual structure compat

piemonkey:
### Overview
Some backwards compatibility fixes for article structure plugin nodetypes.

Requires https://github.com/lblod/ember-rdfa-editor/pull/1128

### Setup
`npm i` should be enough as it's referencing the git commit it needs.

### How to test/reproduce
Paste different old html containing article structures into the 'custom html' input in the editor. HTML can be taken from e.g. qa GN. Check that all the relationships are shown and clicking on outgoing relationships and backlinks works. Refresh and make sure it still works.

### Challenges/uncertainties
Aside from dependency problems, a difficulty was not knowing when something was already broken just being hidden by prosemirror's default behaviour. It would also be very valuable to have a collection of old documents that we can test with.

### Checks PR readiness
- [ ] changelog
- [x] npm lint
- [x] no new deprecations


package.json
```diff
@@ -114,7 +114,7 @@
     "ember-cli-terser": "^4.0.2",
     "ember-cli-typescript": "^5.2.1",
     "ember-disable-prototype-extensions": "^1.1.3",
-    "ember-intl": "5.7.2",
```
abeforgit, line 117: do you need 6.1? afaik we had some issues with it @elpoelma if you could weigh in

piemonkey: Ah, I saw [this commit](https://github.com/lblod/ember-rdfa-editor/pull/1122) moving to 6.1 and assumed this had been done on this branch to avoid adding another breaking release and instead combining it with the rdfa work. I see now that the PR was actually to master and then it was reverted again later. The confusion was that the commit moving back to 5.7 wasn't in `feat/rdfa-links-ui`, so I got unresolved peer dep errors.

I'll merge master into that branch and then rebase this without that commit.

PR #388, opened 2024-02-14
Rdfa/blackbox tests

piemonkey:
### Overview
Blackbox tests.

Currently fails with an issue that I can't reproduce by pasting into rdfa.org/play, which implies that it's a difference in parsing between that and rdfajs:

```
In 'before' but not in 'after':

<http://data.lblod.info/titles/6112780c-ed34-4cea-8bc7-b00286b917d7> <say:heading> "I. This is a Title"@nl-be .
            
In 'after' but not in 'before':

<http://data.lblod.info/titles/6112780c-ed34-4cea-8bc7-b00286b917d7> <say:heading> "I. "@nl-be .
```

Requires https://github.com/lblod/ember-rdfa-editor/pull/1139
Built on https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/387

### Setup
Should all be linked with git version dependency.

### How to test/reproduce
`npm run test:ember -- --launch <browser>`

### Challenges/uncertainties
Type problems when linking the `editor` repo and these more relevant problems:
Copied from standup:
- We don't handle subjects of a namedNode which only exist outside of the document. This seems to be needed for any of the article structures. I've got a branch with a hack that allows this to work, I'll put it in a draft PR but that's just to get the tests which will be in my plugins PR (closer to) passing.
- Our new nodes modify the HTML content, so any content node with datatype XMLLiteral is different after parsing. I fixed this for very basic examples at one point, but there's not really a way that I can think of to fix it for the general case without handling RDFa completely differently. We can also remove this datatype from any blackbox tests to allow those to work.
- We seem to have some problems with nodes not being pre-parsed correctly, which I didn't manage to figure out before deciding I should move on. I seem to be getting elements with RDFa attributes that haven't got the attributes I would expect from the pre-parser (rdfaNodeType, backlinks, etc). This was definitely working when I was last looking at this stuff. This is the last piece that is keeping my blackbox test on plugins failing, despite it just consisting of a title node with the datatype stripped from the body.

### Checks PR readiness
- [ ] changelog
- [ ] npm lint
- [ ] no new deprecations


abeforgit:
I think we just have to accept that we're gonna be mutating xml literals, I don't really see a legitimate way around it either, and I think it's an acceptable compromise

elpoelma:
The failing `title` test indeed seems to be an error with the parser. I also agree with @abeforgit that mutating xml literals is acceptable.

PR #393, opened 2024-02-28
GN-4729: Integrate structure start numbers work with RDFA branch

dkozickis:
GN-4729: Integrate structure start numbers work with RDFA branch

See testing instructions here - https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/376

### DIFF
--- a/addon/plugins/article-structure-plugin/commands/recalculate-structure-numbers.ts
+++ b/addon/plugins/article-structure-plugin/commands/recalculate-structure-numbers.ts
@@ -54,6 +54,14 @@ export default function recalculateStructureNumbers(
     structureSpecs.length,
   ).fill(null);
 
+  /**
+   * Recording all calls to `updateNumber` in an array to run them all
+   * _after_ we've run all the `setNumber` calls.
+   *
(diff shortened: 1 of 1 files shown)


addon/plugins/article-structure-plugin/commands/recalculate-structure-numbers.ts
```diff
@@ -38,30 +49,45 @@ export default function recalculateStructureNumbers(
 ) {
   const doc = transaction.doc;
   const indices = new Array<number>(structureSpecs.length).fill(1);
+
   const contexts: (PNode | null)[] = new Array<PNode | null>(
     structureSpecs.length,
   ).fill(null);
+
+  const updateNumberCalls: UpdateNumberProps[] = [];
```
abeforgit, line 57: bit of a comment here would be nice, it's not the most intuitive thing (though I do think I get why you do it like this)

abeforgit (changes requested):
looks good! and thank you for untangling this :pray:

elpoelma (approved):

PR #397, opened 2024-03-12
Use full uris in decision-type plugin

### LINKED ISSUE #396: The `rdfaAware` version of the decision-type plugin fails due to prefixed vs non-prefixed predicates
The `rdfaAware` version of the decision-type plugin currently fails due to the usage of prefixed vs non-prefixed predicates.

**When setting the decision type, the following happens:**
1. Current decision type is determined by filtering the properties of the decision node
2. `removeProperty` command is executed on current decision type with the prefixed version of `RDF('type')` https://github.com/lblod/ember-rdfa-editor-lblod-plugins/blob/99b660ca9e3acd8fa30b629e1dfa94ef59dab3c9/addon/components/besluit-type-plugin/toolbar-dropdown.ts#L213-L218
3. `addProperty` command is executed with newly selected decision type with the prefixed version of `RDF('type')` https://github.com/lblod/ember-rdfa-editor-lblod-plugins/blob/99b660ca9e3acd8fa30b629e1dfa94ef59dab3c9/addon/components/besluit-type-plugin/toolbar-dropdown.ts#L224-L230

**The problem:**
After reloading a document containing a decision with a decision type, the `RDF('type')` predicate gets expanded due to prefixes.
The `removeProperty` command fails to do something as it can no longer match the prefixed version of `RDF('type')`

I created an issue for this problem, as it is a problem which will probably occur in other plugins/instances too.

**Solutions**
- When inserting a predicate/property, we should always use the 'long' version of the URI, and not the prefixed one
- Long-term: only allow inserting prefixed predicates/uris when those prefixes have been defined in the document.

elpoelma:
### Overview
This PR ensures that the full `RDF('type')` uri is used when inserting/removing decision types from a decision.

Closes #396 

### How to test/reproduce
- `pnpm start`
- Insert a decision
- Select a decision type
- Reload the page
- Select a new decision type
- The old decision type should be correctly removed

### Challenges/uncertainties
- When setting a new decision type, it would probably be better to remove all previous decision type properties (just to be sure). I can still update this PR with this.

### Checks PR readiness
- [ ] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/addon/components/besluit-type-plugin/toolbar-dropdown.ts
+++ b/addon/components/besluit-type-plugin/toolbar-dropdown.ts
@@ -5,7 +5,6 @@ import { addProperty, removeProperty } from '@lblod/ember-rdfa-editor/commands';
 import { SayController } from '@lblod/ember-rdfa-editor';
 import { sayDataFactory } from '@lblod/ember-rdfa-editor/core/say-data-factory';
 import { ResolvedPNode } from '@lblod/ember-rdfa-editor/plugins/datastore';
-import { getRdfaAttribute } from '@lblod/ember-rdfa-editor/utils/rdfa-utils';
 import fetchBesluitTypes, {
   BesluitType,
 } from '@lblod/ember-rdfa-editor-lblod-plugins/plugins/besluit-type-plugin/utils/fetchBesluitTypes';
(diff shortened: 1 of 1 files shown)


piemonkey (approved):
I think removing all existing decision types would be slightly better but also I can see that as being potentially quite messy, so it's probably not worth it.

elpoelma:
> I think removing all existing decision types would be slightly better but also I can see that as being potentially quite messy, so it's probably not worth it.

Yes indeed.

PR #407, opened 2024-04-02
Update to latest AU and use inline svgs

piemonkey:
### Overview
Update to the latest ember-appuniversum so that we can pass inline svgs as icon arguments. Update all usages of appuniversum icons to pass icon components instead of strings. Required some fixing of deprecated uses of AU components.

Includes (almost) the same tooling changes as for the editor, to allow for support of single-file components. This also meant modernising the TS set-up, for example, by removing ember-cli-typescript.

Part of https://github.com/lblod/frontend-embeddable-notule-editor/issues/243
Uses the work on https://github.com/appuniversum/ember-appuniversum/issues/482
And https://github.com/lblod/ember-rdfa-editor/pull/1178

### Setup
You might need to do some set-up in your editor if you haven't set it up to support the new template syntax, details [in the ember docs](https://guides.emberjs.com/release/components/template-tag-format/#toc_editor-integrations) for highlighting and [the glint docs](https://typed-ember.gitbook.io/glint/getting-started#glint-editor-extensions) for type checking.

### How to test/reproduce
All the icons should work. If you inspect an icon, you should no longer see a <use> element inside the <svg>, you should just see the svg inline.

Due to deprecation fixes, should also check variable plugin, citation plugin and snippet plugin for UI problems.

### Challenges/uncertainties
There's quite some built-up technical dept. This fixes some of it, but there are differences between the set-up here and in the editor that are most likely historical and make it hard to make updates in both places at once.

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations


abeforgit (approved):
I can't find a fault so this is good to go for me

PR #408, opened 2024-04-16
Allow to insert structure even if insertion range does not have RDFA attributes.

dkozickis:
### Overview

Allow to insert structure even if insertion range does not have RDFA attributes.

`insertStructure` was too strict checking whether `containerNode` found by `findInsertionRange` has RDFA attributes or not.

### Setup

1. Checkout

### How to test/reproduce

Try inserting 2 of the same structures in a row, before the fix you need to click under the structure you added, add a newline, and then you can insert the next one, after the fix you can spam them all you want.

#### Before

https://github.com/lblod/ember-rdfa-editor-lblod-plugins/assets/769698/634d8253-235b-41b8-a99d-eb4f37c6bd65

#### After

https://github.com/lblod/ember-rdfa-editor-lblod-plugins/assets/769698/9ef10b28-9919-404b-9028-c8e2c0de96e5

### Challenges/uncertainties

My understanding is that it should be possible to insert structure regardless, as structures _can_ be used without RDFA. Is that correct?

### Checks PR readiness
- [X] changelog
- [X] npm lint
- [X] no new deprecations

### DIFF
--- a/package.json
+++ b/package.json
@@ -176,5 +176,6 @@
   },
   "ember-addon": {
     "configPath": "tests/dummy/config"
-  }
+  },
+  "packageManager": "pnpm@8.15.7+sha256.50783dd0fa303852de2dd1557cd4b9f07cb5b018154a6e76d0f40635d6cee019"
 }


abeforgit (approved):
lgtm!

PR #410, opened 2024-04-23
Removal of redundant attributes and parse-rules from decision node-specs

elpoelma:
### Overview
This PR removes redundant attributes and parse-rules from the following node-specs:
- `besluit`
- `besluit_title`
- `description`
- `motivering`
- `article_container`
- `besluit_article`
- `besluit_article_content`

The parse-rules of these nodespecs should be able to parse both the old and the new representations.

None

### How to test/reproduce
- Start the dummy app
- Import the sample `DecisionTemplate` or copy over a document from GN
- Ensure that all RDFa contained in the decision is correctly parsed and stored.

### Challenges/uncertainties
At the moment, there are still issues in the parsing logic of decision-articles, I will tackle these issues in a seperate PR.

### Checks PR readiness
- [ ] changelog
- [x] npm lint

### DIFF
--- a/addon/plugins/standard-template-plugin/utils/nodes.ts
+++ b/addon/plugins/standard-template-plugin/utils/nodes.ts
@@ -491,10 +491,6 @@ export const besluit: NodeSpec = {
   canSplit: false,
   attrs: {
     ...rdfaAttrSpec({ rdfaAware }),
-    typeof: {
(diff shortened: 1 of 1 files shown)


piemonkey (approved):
I can't remember the case that these were added for, but they should be easy to reinstate if we find problems again.

PR #419, opened 2024-04-26
Placeholder for snippets

piemonkey:
### Overview
The title says it all really... I tried to understand the intent of the ticket, so hopefully this makes sense. I've added a node to act as a placeholder. When inserted, we ask which snippet lists should be allowed for that placeholder. It uses the same RDFa as the existing logic to hold snippet list ids.

https://binnenland.atlassian.net/browse/GN-4782
Uses: https://github.com/lblod/ember-rdfa-editor/pull/1188

### Setup
Includes changes from editor, but they are included by using a prerelease dependency.

### How to test/reproduce
Insert a placeholder, then select it and insert a snippet.

### Challenges/uncertainties
...

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations


abeforgit (approved):
That's indeed pretty much exactly what I had in mind. One thing I think we'll need is the ability to edit the text inside the placeholder, to give more informative instructions to the user. But for that we can await saska's design, which will take this requirement into account as well

PR #426, opened 2024-05-09
GN-4692: Create a `besluit-topic` (decision topic) plugin

dkozickis:
### Overview

Create a `besluit-topic` plugin allowing to connect multiple topics (`BesluitThema`) to a decision (besluit).

https://binnenland.atlassian.net/browse/GN-4692

### Setup

1. Checkout

### How to test/reproduce

Please repeat after me, see the video :) 

https://github.com/lblod/ember-rdfa-editor-lblod-plugins/assets/769698/c6b4fd79-aac8-4cbb-a4bc-25a530d47532

### Challenges/uncertainties

1. Plugin provider an UI element that can be inserted anywhere, similar to `besluit-type` plugin. But there's not enough space in demo, or in GN, to insert it without wrapping the menu. So that's kind of meh :( 

2. It's possible to have a sub-topic for a topic (see [here ](https://github.com/Informatievlaanderen/OSLOthema-slimmeRaadpleegOmgeving?tab=readme-ov-file#subthemas-defini%C3%ABren)), but I have not implemented it yet 🤔 @abeforgit should it be done? 

### Checks PR readiness
- [X] changelog
- [X] npm lint
- [X] no new deprecations

### DIFF
--- a/addon/components/besluit-topic-plugin/besluit-topic-toolbar-dropdown.hbs
+++ b/addon/components/besluit-topic-plugin/besluit-topic-toolbar-dropdown.hbs
@@ -83,6 +83,11 @@
           />
         {{/if}}
       </Modal.Body>
+      <Modal.Footer>
+        <AuButton {{on 'click' this.toggleCard}}>{{t
+            'besluit-topic-plugin.close-modal'
+          }}</AuButton>
+      </Modal.Footer>
     </AuModal>
   {{/if}}
 </div>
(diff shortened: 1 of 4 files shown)


abeforgit (changes requested):
in general looks good, just some UX concerns:
- afaik topic is not a mandatory field, so we shouldn't throw the modal in people's faces on load
- because it's a multiselect, we shouldn't close the modal after selecting one. A modal that stays open, with an "ok" button that just closes the modal (cause it autosaves on select already) would work better I think

lagartoverde (approved):
In my opinion the pill should have a pointer to show its clickable but the decision type one doesn't have it either, so I guess it's okay, everything else works fine :)

PR #427, opened 2024-05-13
Feature/lmb plugin

lagartoverde:
### Overview
Added a new plugin to add info about mandatees on LMB to the document, this will need the service lmb-proxy-service in the backend, also you should be logged to an admin unit

https://binnenland.atlassian.net/browse/GN-4830
https://github.com/lblod/app-gelinkt-notuleren/pull/182

### Setup
Clone the backend branch, build the service, link the plugin to the frontend and add it to the document

### How to test/reproduce
Click on the side menu, add a new lmb mandatee and that's it

### Challenges/uncertainties
I used the css of  the worship service, as we are waiting for design on this plugin, should we wait until we do, or should I just convert those clases to neutral naming?

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations


abeforgit (changes requested):
The types aren't correct:
```
addon/components/lmb-plugin/search-modal.ts:95:9 - error TS2322: Type 'Mandatee[] | null | undefined' is not assignable to type 'Mandatee[]'.
  Type 'undefined' is not assignable to type 'Mandatee[]'.

95     let mandatees: Mandatee[] = this.fetchData.lastComplete?.value;
           ~~~~~~~~~

addon/components/lmb-plugin/search-modal.ts:105:30 - error TS2339: Property 'toSorted' does not exist on type 'Mandatee[]'.

105       mandatees = mandatees?.toSorted((a: Mandatee, b: Mandatee) => {
```

addon/components/lmb-plugin/search-modal.ts
```diff
@@ -0,0 +1,157 @@
...
+        `,
+      }),
+    });
+    const queryJson: SparqlResponse = await queryResponse.json();
+    const mandatees = queryJson.results.bindings.map(Mandatee.fromBinding);
+    return mandatees;
+  });
+
+  search = restartableTask(async () => {
+    // Can't do what I want, so if the user modifies the filter before resolving the query will run again
```
abeforgit, line 90: what did you want to do?

addon/components/lmb-plugin/list.hbs
```diff
@@ -0,0 +1,60 @@
+<AuTable>
+  <:header>
+    <tr class='au-c-data-table__header-title'>
+      <th>
+        <WorshipPlugin::SortableTableHeader
```
abeforgit, line 5: If you're reusing these, it makes sense to extract them out from under the worshipPlugin namespace

translations/en-US.yaml
```diff
@@ -416,3 +416,17 @@ worship-plugin:
       sort: Sort
       asc: Sort ascending
       desc: Sort descending
+
+lmb-plugin:
```
abeforgit, line 420: missing dutch translations

abeforgit (changes requested):
haven't managed to get it running, but here's already some code comments

abeforgit (changes requested):
some more things:
- name sorting should be on the last name, if possible
- name search should be case-insensitive

addon/components/lmb-plugin/search-modal.ts
lagartoverde: Basically I wanted to avoid refetching the data from the endpoint if the user changes the sort or filter before it finishes, but I couldn't I don't think is really relevant but would be nice to do it.
I don't seem to know enough about ember-concurrency to do what I want

lagartoverde: As the filter is only applied on the frontend there's no need of refetching the data

addon/components/lmb-plugin/list.hbs
lagartoverde: I used Common::Sort::SortableTableHeader is this a good naming?

translations/nl-BE.yaml
```diff
@@ -412,7 +416,18 @@ worship-plugin:
...
+  modal:
+    title: Kies een mandataris
+    insert: Invoegen
+    search:
+      title: Filter op
+    fields:
+      name: Naam
+      status: Status
+      fractie: Fractie
+      role: Role
```
abeforgit, line 433: role => "Rol"

abeforgit (changes requested):
Just one more translation thing, you don't need to re-request review after changing it, just merge

PR #428, opened 2024-05-17
Style exported snippet placeholders, and fix styling for inline placeholders for full URIs

lagartoverde:
### Overview
Implemented some stylng fixes related to GN-4826 and GN-4832
GN-4832
GN-4826

### How to test/reproduce
Link the plugins to the frontend and go to regulatory-statements/6644C5FDD4D58A6BA3A6C75E/edit while being proxied to dev, in Geemente Aalst, it should show correctly.
You can also add this same RB to an agendapoint, and it should show as a mark highlight

### Challenges/uncertainties
Decided to go with the span in the end, because the div resulted in a lot of copy pasted css, and had some problems with it being a div. If we get some design we can change to its own css class and style accordingly

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations


abeforgit (changes requested):
I don't like how the snippet placeholder now looks exactly like other text placeholders... 
We talked about them having a similar background, not using the exact same class

And I thought we agreed on giving it its own class, and keeping it as a div...

abeforgit (approved):

PR #429, opened 2024-05-17
GN-4693: Insert LPDC

dkozickis:
### Overview

Insert LPDC reference for besluit (decision)

* https://binnenland.atlassian.net/browse/GN-4693
* https://github.com/lblod/app-gelinkt-notuleren/pull/183
* https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/429

### Setup

1. Checkout frontend PR - https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/429
2. Checkout `app-gelinkt-notuleren` PR - https://github.com/lblod/app-gelinkt-notuleren/pull/183
3. Start `app-gn` with something like `docker compose -f docker-compose.yml -f docker-compose.dev.yml -f docker-compose.override.yml up`, `API_KEY` for `lpdc-service` should go into `docker-compose.override.yml`
4. Start frontend

### How to test/reproduce

Repeat after me

https://github.com/lblod/ember-rdfa-editor-lblod-plugins/assets/769698/52f93371-8fa2-4950-92af-58aa985d6600

### Challenges/uncertainties

* What needs to be shown in the search modal?
* How should the node look in the UI really?

### Checks PR readiness
- [X] changelog
- [X] npm lint
- [X] no new deprecations

### DIFF
--- a/tests/dummy/app/controllers/besluit-sample.ts
+++ b/tests/dummy/app/controllers/besluit-sample.ts
@@ -308,7 +308,7 @@ export default class BesluitSampleController extends Controller {
         endpoint: 'http://localhost/vendor-proxy/query',
       },
       lpdc: {
-        // endpoint: 'https://dev.gelinkt-notuleren.lblod.info/lpdc-service',
+        // Needs to point at the same port as the ember app
         endpoint: 'http://localhost/lpdc-service',
       },
     };


addon/components/lpdc-plugin/lpdc-insert.ts
```diff
@@ -0,0 +1,94 @@
...
+    if (!resource) {
+      throw new Error('No besluit found in selection');
+    }
+
+    this.controller.withTransaction(
+      (tr) => {
+        const node = this.controller.schema.node(
+          'inline_rdfa',
+          {
+            rdfaNodeType: 'literal',
```
piemonkey, line 70: The way I read [the spec](https://github.com/Informatievlaanderen/OSLOthema-slimmeRaadpleegOmgeving#lpdc-codes) is that this should be a resource. I'm guessing since it isn't clear where the `dcterms:source` in their example comes from, this is only necessary if we want to add extra triples to the otherwise externally defined resource? But literal seems wrong according to their example.

addon/plugins/lpdc-plugin/api.ts
```diff
@@ -0,0 +1,111 @@
...
+      Accept: 'application/json',
+    },
+  });
+
+  const resultJson = (await results.json()) as FetchResults;
+
+  return {
+    lpdc: (resultJson.hydraMember ?? []).map((lpdc) => ({
+      uri: lpdc['@id'],
+      name: lpdc.naam.nl,
```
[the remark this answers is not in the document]
piemonkey: Hmm, filtering seems an odd choice, even though I agree that we can probably assume everything has `nl`. I'd personally probably do `naam.nl ?? naam.en` and filter out any that still have nothing.

addon/components/lpdc-plugin/lpdc-insert.ts
```diff
@@ -0,0 +1,93 @@
...
+      },
+      { view: this.controller.mainEditorView },
+    );
+
+    this.controller.doCommand(
+      addProperty({
+        resource,
+        property: {
+          predicate: SRO('bekrachtigt').full,
+          object: sayDataFactory.literalNode(rdfaId),
```
piemonkey, line 86: Should this be `.resourceNode(uri)`?

piemonkey (approved):

PR #430, opened 2024-05-17
Use OSLO model for addresses

piemonkey:
### Overview
Tweak the representation of addresses produced by variable-plugin to conform to the OSLO model.

Includes converting GML location data to WKT as that is used in the spec. The conversion used only works for simple points, but we don't handle anything but this already. Notes as to why I used a regex instead of a full parser are in comments.

https://binnenland.atlassian.net/browse/GN-4838

### Setup
N/A

### How to test/reproduce
Inserting and editing addresses should work correctly. The exported HTML should match the OSLO data model for addresses.

### Challenges/uncertainties
We slightly abuse the model by re-using the id of the address as the resource id, which their examples do not. I assumed this was fine though.
I reverted my changes to use language correctly as it caused too many issues to be worth it, given everything will be in 'nl' (even the name of the field returned by the API is 'taal').
~~I also haven't converted the location point to use WKT, which is implied as being required in their docs. We could easily add a module to handle conversions from GML and do that however.~~

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/nervous-paws-yawn.md
+++ b/.changeset/nervous-paws-yawn.md
@@ -1,0 +1,5 @@
+---
+"@lblod/ember-rdfa-editor-lblod-plugins": minor
+---
+
+Use OSLO model for addresses in variable-plugin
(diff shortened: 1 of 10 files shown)


abeforgit (approved):
Works absolutely beautifully, including parsing and converting old addresses, great job!

PR #432, opened 2024-05-23
Change location-plugin to use a modal to edit addresses

piemonkey:
### Overview
Much of the code is just moved around but some changes were necessary to handle the state in different parts of the structure.

This PR is to the branch for just separating the code, so that this PR has changes more than just what looks like new code.

Issue: [binnenland.atlassian.net/browse/GN-4839](https://binnenland.atlassian.net/browse/GN-4839)
PR for splitting out location-plugin: #431 

### Setup
N/A

### How to test/reproduce
Only used on besluit-sample so far. There is an 'insert address' button in the same place as before but it now opens a modal.

### Challenges/uncertainties
I wasn't sure of the UX of this. It feels more awkward than before...

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations


abeforgit (approved):
actually don't mind the UX myself, it's more clicky, but it works better on smaller screens and there's less scrolling problems :man_shrugging: 
including a leaflet in the modal might make the extra click feel more worth it too

but if you have ideas on how we could improve the ux, definitely write them out. The suggestions I make in the tickets are really just the first thing that pops in my mind, which usually means its just a pattern we already have.

I think it's safe to say our role as devs in the GN team extends into design, ux and even analysis

PR #433, opened 2024-05-30
Remove label when variable is filled

lagartoverde:
### Overview
I made so the label is not shown when you fill a variable

GN-4824

### How to test/reproduce
You can insert one of each variables and fill it, appreciate that the label is indeed hidden, in some variables you can go back to an unfilled status, and the variable should reappear, but on others this is just not possible because of how the variable UI is made

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/big-pugs-sneeze.md
+++ b/.changeset/big-pugs-sneeze.md
@@ -1,0 +1,5 @@
+---
+"@lblod/ember-rdfa-editor-lblod-plugins": minor
+---
+
+Hide label if the variable is filled
--- a/addon/components/location-plugin/nodeview.ts
+++ b/addon/components/location-plugin/nodeview.ts
@@ -44,7 +44,7 @@ export default class AddressNodeviewComponent extends Component<Args> {
   }
 
   get label() {
-    if(this.address) return ''
+    if (this.address) return '';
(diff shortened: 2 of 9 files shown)


lagartoverde:
The thinner version works with all of them, and I think is prefered, what I did now is to make it consistent, as some variable were 25px and some 29px, so I set all of them at 30px.
So we will have with label 35px and without 30px, I don't know is this is correct or I should make them all 35. Maybe @abeforgit can throw some light

elpoelma:
> The thinner version works with all of them, and I think is prefered, what I did now is to make it consistent, as some variable were 25px and some 29px, so I set all of them at 30px. So we will have with label 35px and without 30px, I don't know is this is correct or I should make them all 35. Maybe @abeforgit can throw some light

IMO we should probably use the same height whether a label is shown or not. This shows the user that they are handling the same type of data.

piemonkey:
For me, I think thinner is better, as it reduces the impact of adding a variable to the rest of the layout. I think the most important thing is that they don't change size though, as it means the layout changes just by adding content, which to me would indicate that the type of node is changing.

lagartoverde:
You can review this again

piemonkey (approved):

PR #436, opened 2024-06-06
Added the lists of the snippet placeholder in the prosemirror node

lagartoverde:
### Overview
Changed the UI of the snippet placeholder to specify which lists you can insert from.

GN-4852

### How to test/reproduce
Insert a snippet placeholder with only a snippet list and check that it shows as described in the ticket, then change it to multiple lists, it should change the UI to show the multiple lists you can insert from 

### Challenges/uncertainties
None

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/addon/components/snippet-plugin/snippet-list-select-rdfa.ts
+++ b/addon/components/snippet-plugin/snippet-list-select-rdfa.ts
@@ -49,8 +49,11 @@ export default class SnippetListSelectRdfaComponent extends Component<Args> {
     );
   }
 
-  saveChanges = (snippetIds: string[]) => {
+  saveChanges = (snippetIds: string[], snippetNames: string[]) => {
     if (this.currentResource) {
+      this.controller.withTransaction((tr) => {
+        return tr.setNodeAttribute(this.node.pos, 'listNames', snippetNames);
+      });
       this.args.controller?.doCommand(
         updateSnippetIds({
           resource: this.currentResource,
--- a/addon/components/snippet-plugin/snippet-list-select-rdfa.ts
+++ b/addon/components/snippet-plugin/snippet-list-select-rdfa.ts
@@ -7,7 +7,7 @@ import {
   getAssignedSnippetListsIdsFromProperties,
   getSnippetListIdsProperties,
 } from '@lblod/ember-rdfa-editor-lblod-plugins/plugins/snippet-plugin/utils/rdfa-predicate';
-import { updateSnippetIds } from '@lblod/ember-rdfa-editor-lblod-plugins/plugins/snippet-plugin/commands';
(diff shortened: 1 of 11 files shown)


addon/components/snippet-plugin/snippet-list/snippet-list-modal.ts
```diff
@@ -50,7 +50,16 @@ export default class SnippetListModalComponent extends Component<Signature> {
 
   @action
   saveAndClose() {
-    this.args.onSaveSnippetListIds(this.assignedSnippetListsIds);
+    const snippetListNames: string[] = this.snippetListResource.value
+      ?.filter((snippetList) =>
+        this.assignedSnippetListsIds.includes(snippetList.id as string),
+      )
+      .map((snippetList) => snippetList.label) as string[];
+    console.log(snippetListNames);
```
piemonkey, line 58: Console.log

addon/components/snippet-plugin/snippet-list-select-rdfa.ts
```diff
@@ -49,8 +49,11 @@ export default class SnippetListSelectRdfaComponent extends Component<Args> {
     );
   }
 
-  saveChanges = (snippetIds: string[]) => {
+  saveChanges = (snippetIds: string[], snippetNames: string[]) => {
     if (this.currentResource) {
+      this.controller.withTransaction((tr) => {
```
piemonkey, line 54: This removes the name in a separate transaction to the update of the list ids, which puts it in it's own entry in the undo stack. You can see this by modifying the snippet lists assigned to a placeholder, then hitting ctrl-z (you need to de-select the node for some reason, this is most likely an ember-node bug); you should see the same list of names in the node (as after the change) but if you hit 'manage snippet lists' you'll see the original selections from before the change.

I think we need to modify the `updateSnippetIds` command.

translations/nl-BE.yaml
```diff
@@ -273,8 +273,10 @@ snippet-plugin:
     title: Fragment invoegen
     placeholder: Plaatshouder voor fragment invoegen
   placeholder:
-    title: Fragment hier invoegen
-    text: Selecteer dit en voeg het gewenste fragment in
+    title-single: Fragment uit de lijst "{listName}"
+    title-multiple: Fragment uit meerdere lijsten
+    text: Selecteer dit en voeg het gewenste fragment in"
```
piemonkey, line 278: There looks to be a trailing `"` here.

piemonkey (changes requested):
Mostly looks good, just a small bug with undo.

addon/components/snippet-plugin/snippet-list-select-rdfa.ts
lagartoverde: Didn't think of the undo command, thanks, should be fixed now

piemonkey (approved):

PR #439, opened 2024-06-10
Add map and ability to pick points to location plugin

piemonkey:
### Overview
The state management in this is a bit of a mess. It probably would have been easier as one massive component but this way does at least separate out concerns a little better. The async conversion of coordinates and configurable URIs made things more tricky. I'm not sure about the UX of how to indicate that when you're specifying a 'place' the address form is just to help you find things.

https://binnenland.atlassian.net/browse/GN-4840

### Setup
N/A

### How to test/reproduce
On besluit sample, insert, edit and otherwise mess around with address and place locations.

### Challenges/uncertainties
State management and not knowing what the UX should be.

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations


dkozickis:
> I'm not sure about the UX of how to indicate that when you're specifying a 'place' the address form is just to help you find things.

Maybe something like this.

CleanShot 2024-06-12 at 17 06 32@2x

dkozickis:
After switching from "Place" to "Address" input the "Place" marker stays visible. Not sure if a feature or a bug 🙏🏻

addon/components/location-plugin/map.gts
```diff
@@ -0,0 +1,131 @@
...
+        </AuLoader>
+      {{else if this.addressLatLng.value}}
+        <LeafletMap
+          @center={{this.mapCenter}}
+          @zoom={{this.zoom}}
+          @onClick={{this.onMapClick}}
+          as |layers|
+        >
+          <layers.tile
+            @url='https://{s}.tile.openstreetmap.fr/hot/{z}/{x}/{y}.png'
```
dkozickis, line 100: _Non-blocking_: why the "[Humanitarian map style](https://wiki.openstreetmap.org/wiki/Humanitarian_map_style)" instead of the regular one `https://tile.openstreetmap.org/{zoom}/{x}/{y}.png`?

dkozickis: P.S. Also using OSM tiles requires license attribution, right now it only shows the "Leaflet" link. Not sure how easily is that done with `ember-leaflet` 

CleanShot 2024-06-12 at 17 27 37@2x

dkozickis (approved):

piemonkey:
> After switching from "Place" to "Address" input the "Place" marker stays visible. Not sure if a feature or a bug 🙏🏻

It's a bug. Weirdly that video seems to have an additional bug that I can't reproduce, the pin already showing when you loaded the modal...

addon/components/location-plugin/map.gts
piemonkey: Seemed cleaner... Maybe regular OSM is better for contrast though.

abeforgit (changes requested):
image

abeforgit:
few ux nits, but great work overall :1st_place_medal:

abeforgit:
Note I can predict a desired two-way select on the address page (click on map, have address filled in)
but from experience that's a state management hellhole that it's best to put out of scope for now

piemonkey:
> image

Hmm, so when you say 'less of a form', you mean to have different UX for the address and for the place searches? That feels like a lot of extra work for something that we don't have a design for and will likely need to change in the future... It also depends a little on the API, as I'm not sure if it's set up in a way that would make this kind of search easy (I had the impression that it was, but when I just went to check, I couldn't find it. Their documentation isn't the clearest about how inputs are handled though).

I'm not sure what you mean about 2-way binding, the address doesn't update when moving the map, so surely there's no impression of two-way binding?

> "Centreer de kaart rond adres" = "Center the map around address"
> 
> Ah and I read over deniss's suggestion about this, I really like the explanation so I'd add it as well.
> 
> Here's a dutch translation: "Gebruik de onderstaande zoekfunctie om de kaart te centreren rond de gewenste plaats, en klik daarna op de kaart om de locatie te kiezen"

I'll try to add these in a way that isn't overloading with too much text...

piemonkey:
> Note I can predict a request for two-way select on the address page (click on map, have address filled in) but from experience that's a state management hellhole that it's best to put out of scope for now

The geo api seems well set up for this. We'd be relying on them to give us good suggestions, but I actually think this wouldn't be too hard with the current set-up. I think I dealt with the state management hellhole as it is...

abeforgit:
I'd rather have too much text than too little, but there is an example of an info popup in the date editor I believe, so maybe you can use that

"less of a form" was a bit of a shot in the dark, don't really know how to do that. The only thing is to maybe get rid of the mailbox number input, as that's a bit useless for the purposes of centering the map

I had in mind to reshuffle things a bit to make it 'feel' more like a search rather than an input form, but honestly I don't really know what that would look like so ignore that.

With 2-way binding I meant exactly that, to me it "feels" like the address _should_ update when I click on the map, and the fact it doesn't feels like a bug. That's why I wanted to make it more clear that unlike the address form, the address part is _not_ part of the data you're entering and merely a tool to make the map easier to use. The fact the state carries over from the address input and the fact it looks exactly the same strengthens the idea that the address part still matters for more than just the view I guess.

We're getting into design waters here but maybe even a simple thing like adding a background grey or smthing to set them apart? 

I'll leave it to you, the added text should be enough as a baseline

abeforgit:
> > Note I can predict a request for two-way select on the address page (click on map, have address filled in) but from experience that's a state management hellhole that it's best to put out of scope for now
> 
> The geo api seems well set up for this. We'd be relying on them to give us good suggestions, but I actually think this wouldn't be too hard with the current set-up. I think I dealt with the state management hellhole as it is...

I made a similar component once, and I found it hard to get a good ux going where you have 2 different input methods (the map and the form) 
I was inexperienced with frontend work back then though so I may be overestimating it

piemonkey:
> > > Note I can predict a request for two-way select on the address page (click on map, have address filled in) but from experience that's a state management hellhole that it's best to put out of scope for now
> > 
> > 
> > The geo api seems well set up for this. We'd be relying on them to give us good suggestions, but I actually think this wouldn't be too hard with the current set-up. I think I dealt with the state management hellhole as it is...
> 
> I made a similar component once, and I found it hard to get a good ux going where you have 2 different input methods (the map and the form) I was inexperienced with frontend work back then though so I may be overestimating it

What I mean is that right now both the map and the form are updating state on the parent component, so this thing that is indeed hard, has been done. There might well be a gotya hidden in there, but it might just work...

abeforgit:
to explain a bit more what I had in mind (which disregards time and technical constraints completely)

Address form:

as-is, leaflet could even be a bit smaller to reduce significance (it's only there to help you fill in the form)

Location form:

map takes up the whole width of the modal. The leaflet has a search widget like on the osm site (it's only there to help you control the map):
image

Somewhere in unobtrusive small font, there's a display of the selected coordinates to give the confidence that something happens when you click, but not so in-your-face that users get scared seeing weird numbers.

However: don't actually do this, it's completely fine as-is, but I hope this makes it a bit more clear what I mean

piemonkey:
> map takes up the whole width of the modal. The leaflet has a search widget like on the osm site (it's only there to help you control the map)
> 
> Somewhere in unobtrusive small font, there's a display of the selected coordinates to give the confidence that something happens when you click, but not so in-your-face that users get scared seeing weird numbers.

We also need to enter the name for a 'place' so that there's something human-facing in the editor.

abeforgit:
> > map takes up the whole width of the modal. The leaflet has a search widget like on the osm site (it's only there to help you control the map)
> > Somewhere in unobtrusive small font, there's a display of the selected coordinates to give the confidence that something happens when you click, but not so in-your-face that users get scared seeing weird numbers.
> 
> We also need to enter the name for a 'place' so that there's something human-facing in the editor.

yes indeed, forgot to mention that

(but like I said, its ok as-is with just some extra text clarification)

piemonkey:
image
Comments?

abeforgit:
> image Comments?

looks good! ok to merge for me

PR #443, opened 2024-06-21
GN-4841 Add ability to specify area locations in the location-plugin

piemonkey:
### Overview
Required some work to shield ember-leaflet from being confused by args and some type gymnastics. I'm pretty happy with the UX for something simple, but maybe it's not intuitive enough? I'll try to get leaflet-draw to work.

https://binnenland.atlassian.net/browse/GN-4841

### Setup
N/A

### How to test/reproduce
On the besluit sample, 'insert location', then select area. I'm not going to say how to use it, as it'll help you see if it's intuitive enough. Once you've figured out how to do it, try making different location types, hoping between editing different ones, refreshing, etc.

### Challenges/uncertainties
Still not sure what we want to do on the template builder side as we can't set a label yet, so it's still using the old address plugin.

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations


abeforgit (changes requested):
Overall works really well, they'll love this

Theres one quirk that's a bit annoying though:

- make a shape
- change the zoom or positioning
- make a new shape
on completing the new shape, the map snaps to the zoom and positioning you used for the first shape.

abeforgit:
and I think its very intuitive

piemonkey:
I think all the jumping around style issues should be fixed now.

> (unsure if already previously discussed) do we want to open the edit-location modal automatically after inserting a location-placeholder?

This hasn't been discussed, but it seemed natural to me, since this is not (at least currently) for making templates, as there is no way to set a label for the placeholder. So the only reason to insert a placeholder is because you want to insert a location immediately. It's easy to change if we think it makes more sense to not auto-open.

elpoelma:
> I think all the jumping around style issues should be fixed now.
> 
> > (unsure if already previously discussed) do we want to open the edit-location modal automatically after inserting a location-placeholder?
> 
> This hasn't been discussed, but it seemed natural to me, since this is not (at least currently) for making templates, as there is no way to set a label for the placeholder. So the only reason to insert a placeholder is because you want to insert a location immediately. It's easy to change if we think it makes more sense to not auto-open.

Yeah that makes sense. If we would still want to change this behaviour in the future, we can always do it in a seperate PR.

elpoelma (changes requested):
The issue with the zoom-out when drawing an area seems to be fixed now :smile: . State is also correctly reset. 
There still seems to be an issue when editing the location: after drawing a new polygon, the map moves to the location where the previous polygon was.

piemonkey:
Okay, let's try that again... I think everything should be fixed now in terms of the map jumping around unexpectedly.

elpoelma (approved):
Looks good now! Really cool!

abeforgit:
weighing in on the "open the modal automatically": it didn't even register for me as that's exactly what I was expecting to happen, so I think we're good

PR #444, opened 2024-06-28
GN-4897: ensure older decision articles can be correctly parsed into the new `structure` nodes

elpoelma:
### Overview
This PR includes the following changes:
- A fix for the structure number serialization
- Disabling the editable title for structures representing decision articles
- Addition of a backwards-compatible parsing-rule to the `structure` node-spec, responsible for parsing decision articles in the old format.

[GN-4897](https://binnenland.atlassian.net/browse/GN-4897)

### How to test/reproduce
- Open the dummy app
- Insert a sample decision template (which at the moment still contains decision articles in the old format)
- Ensure the decision (include the articles in the new format) are correctly parsed
- When adding new article structures, the numbering should still hold up

### Challenges/uncertainties
Note: the backwards compatible parsing-rule throws away the RDFa it parses from the old articles. As the old articles were not yet used in an rdfa-aware way, this should not be a problem.

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations


elpoelma:
@abeforgit I updated this PR with the adjustments to the decision-article RDFa we talked about.
- It adds a `PROV('value')` content-literal
- The ELI('number') is done in a bit of a different way: the simple `number` attribute is still used, the `serialize` method adds the number-RDFa, and the parseDOM filters it out. This ensures we do not need to operate on the RDFa in the editor itself. If you think it is a better idea to store the number as RDFa in the properties, we can still do that of course.

abeforgit (approved):

PR #445, opened 2024-06-28
Gn 4877

lagartoverde:
### Overview
Implemented Snippet fragments per Saska design, changed a little bit the text of the add new button, but otherwise the rest should be the same as the figma

GN-4877

### Setup
None

### How to test/reproduce
Insert a snippet placeholder, insert a snippet and then select it, you should see the buttons

### Challenges/uncertainties
I experienced some problems sometimes with clicking on the buttons, but then Elena couldn't reproduce them, so not sure if its my setup or not, looking for feedback on that so please test extensively

### Checks PR readiness
- [ ] changelog
- [ ] npm lint
- [ ] no new deprecations

### DIFF
--- a/addon/components/snippet-plugin/snippet-insert.ts
+++ b/addon/components/snippet-plugin/snippet-insert.ts
@@ -66,11 +66,13 @@ export default class SnippetInsertComponent extends Component<Args> {
     if (documentDiv) {
       return this.controller.withTransaction((tr: Transaction) => {
         return tr.replaceSelectionWith(
-          htmlToDoc(content, {
-            schema: this.controller.schema,
-            parser,
-            editorView: this.controller.mainEditorView,
-          }),
+          this.controller.schema.node('snippet', {assignedSnippetListsIds: this.args.assignedSnippetListsIds},
+            htmlToDoc(content, {
+              schema: this.controller.schema,
+              parser,
+              editorView: this.controller.mainEditorView,
+            }).content
+          ),
         );
       });
     }
--- a/addon/components/snippet-plugin/snippet-insert.ts
+++ b/addon/components/snippet-plugin/snippet-insert.ts
@@ -55,7 +55,7 @@ export default class SnippetInsertComponent extends Component<Args> {
   }
 
   @action
-  onInsert(content: string) {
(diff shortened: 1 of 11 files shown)


elpoelma (changes requested):
Seems to be working well overall! Nice work :)
I just have some comments in terms of UX and some bugs:
- The gap-cursor around snippet nodes does not seem to be working. This is probably a config issue. It seems to be working with template comments.
- I would increase the size of the buttons a bit, especially the help-text seems pretty small.
- When entering twice inside a snippet (without typing any text), the user is moved outside of the snippet
- When the snippet is filled with text from left to right, the text sometimes interferes with the buttons and help text. 
Screenshot from 2024-06-28 17-41-04
- When the snippet node is very short, the buttons do not fit inside.

addon/components/snippet-plugin/snippet-insert.ts
```diff
@@ -66,11 +67,19 @@ export default class SnippetInsertComponent extends Component<Args> {
...
-            schema: this.controller.schema,
-            parser,
-            editorView: this.controller.mainEditorView,
-          }),
+          this.controller.schema.node(
+            'snippet',
+            {
+              assignedSnippetListsIds: this.args.assignedSnippetListsIds,
+              title: title,
+              subject: `http://example.net/lblod-snippet/${uuidv4()}`,
```
elpoelma, line 75: We should probably use a different URI for this.

addon/plugins/snippet-plugin/nodes/snippet.ts
```diff
@@ -0,0 +1,88 @@
...
+  },
+  component: SnippetComponent,
+  content: '(block | title | chapter | article)*',
+  serialize(node) {
+    return renderRdfaAware({
+      renderable: node,
+      tag: 'div',
+      attrs: {
+        ...node.attrs,
+        class: 'say-snippet-placeholder-node',
```
elpoelma, line 52: Is this class necessary here?

elpoelma (commented):
Looks good now! Styling and UX is intuitive and works well functionality-wise.
I have just 2 questions.

PR #448, opened 2024-07-03
Upgrade many dependencies - breaking

piemonkey:
### Overview
As https://github.com/lblod/ember-rdfa-editor/pull/1204, since the ember-template-imports upgrade required other breaking updates, update a lot of dependencies.

Drops support for ember-intl v5 and v6. In v7 the ember-intl api is simplified, with no version including both removed functions and the new ones, so we either need to do a completely breaking release or use embroider macros to support both.

Uses https://github.com/lblod/ember-rdfa-editor/pull/1204

### Setup
N/A

### How to test/reproduce
Everything should just work.

### Challenges/uncertainties
See editor PR

### Checks PR readiness
- [ ] changelog
- [ ] npm lint
- [ ] no new deprecations


elpoelma:
When adding an article structure to a decision, I'm now getting the following (ember-modifier) error: 
image

piemonkey:
@elpoelma Good spot, looks like it was wrapping the method in `modifier()` twice by accident. Not sure why that is only a problem in the newer ember-modifier, but it should be fixed now.

package.json
```diff
@@ -102,108 +102,117 @@
...
     "leaflet": "^1.9.4"
   },
-  "overrides": {
-    "ember-intl": {
-      "typescript": "^5.0.4"
+  "peerDependenciesMeta": {
+    "@glint/template": {
+      "optional": true
+    },
+    "ember-template-imports": {
```
abeforgit, line 208: glint I get, but is e-t-i really optional? Doesn't it provide the build hooks to compile the gts files? remember this is still a v1 addon and as such is "compiled" by the app at app build time

piemonkey: It's not really optional. An optional peer dep seems to be the only way to say "if you install this, it needs to satisfy these version constraints". Since it's still in the dependencies rather than being moved to dev dependencies, it'll always be installed, in npm/yarn *somewhere* in node_modules and in pnpm nested under lblod/plugins. It does feel weird to put it here, but since there's no way to write comments in the package.json, I can't explain it in the file...

elpoelma: Yeah it's weird indeed, I guess this is why https://github.com/salsify/ember-cli-dependency-lint was created.

abeforgit: "if you install this, it needs to satisfy these version constraints" -> I was under the impression that this is what peerdeps did already? 

you'd remove it from the deps and only add it peerdeps in this case.

abeforgit: yeah but I thought that's what peerdeps are for

you add it to devdeps and peerdeps so:
- client does not depend: its automatically installed by any version of npm we still support
- client does depend: version clashes are errors, and if no clash, the client's version is used
- building the test app: the devdep is used

abeforgit: note that's my understanding of how peerdeps work, I'm fully open to the possibility I'm wrong

piemonkey: The choice of deps rather than peerdeps is down to Ember (I assume just for v1 addons). It treats production dependencies differently to dev dependencies. Simply moving e-t-i to devdeps and leaving nothing else unchanged leads to a runtime error when trying to use a component defined in a gts file. Otherwise, I agree that it should be a devdep as it should only be needed at compile time.

abeforgit: oh wow yeah hadn't considered ember throwing a wrench in the works

PR #449, opened 2024-07-04
Drop obsolete decision nodes from standard-template plugin

elpoelma:
### Overview
This PR removes the obsolete decision nodes from the `standard-template` plugin. The nodes have been replaced by generic `block_rdfa` nodes, and `structure` nodes.
This is of course, a breaking change.

### How to test/reproduce
The test-app no longer relies on these decision nodes, so nothing should have changed.

### Challenges/uncertainties
We could, of course also mark these nodes as deprecated, in order to prevent a major release. This is, however, not as important for this package.

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations


abeforgit (approved):
looks good, but I'm assuming another pr to make the ST plugin compatible with block-rdfa is still on the way?

elpoelma:
> looks good, but I'm assuming another pr to make the ST plugin compatible with block-rdfa is still on the way?

It already works with `block_rdfa` nodes no? It doesn't seem it needs any adaptation?

abeforgit:
aha hmm it must have been breaking for some other reason in GN then, will investigate

PR #454, opened 2024-07-11
Switch some name based icons to use AU icon components

piemonkey:
### Overview
I spotted some of these in my adventures through the codebase, so thought to change them. I also spotted that my comment about template only components that wasn't necessary (or true), so removed it, and added the eslint config to warn on unused ignores, as well as adding a single demo of a typed template only component.

N/A

### Setup
N/A

### How to test/reproduce
Everything still works!

### Challenges/uncertainties
N/A

### Checks PR readiness
- [ ] changelog
- [ ] npm lint
- [ ] no new deprecations

### DIFF
--- a/addon/components/confidentiality-plugin/redact.gts
+++ b/addon/components/confidentiality-plugin/redact.gts
@@ -1,13 +1,20 @@
-import Mark from '@lblod/ember-rdfa-editor/components/toolbar/mark';
+import { TemplateOnlyComponent } from '@ember/component/template-only';
 import t from 'ember-intl/helpers/t';
 import { NotVisibleIcon } from '@appuniversum/ember-appuniversum/components/icons/not-visible';
+import { SayController } from '@lblod/ember-rdfa-editor';
+import Mark from '@lblod/ember-rdfa-editor/components/toolbar/mark';
 
-<template>
+interface Sig {
+  Args: {
+    controller: SayController;
+  };
+}
+
+export const redact: TemplateOnlyComponent<Sig> = <template>
   <Mark
     @icon={{NotVisibleIcon}}
     @title={{t 'confidentiality-plugin.redact'}}
     @mark='redacted'
-    {{! @glint-expect-error: not typesafe yet }}
     @controller={{@controller}}
   />
-</template>
+</template>;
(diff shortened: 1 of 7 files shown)


addon/components/confidentiality-plugin/redact.gts
```diff
@@ -1,13 +1,20 @@
...
+import Mark from '@lblod/ember-rdfa-editor/components/toolbar/mark';
+
+interface Sig {
+  Args: {
+    controller: SayController;
+  };
+}
 
-<template>
+export const redact: TemplateOnlyComponent<Sig> = <template>
```
abeforgit, line 13: Funnily enough, I'm now getting errors with this one:
```
Uncaught Error: Attempted to load a component, but there wasn't a component manager associated with the definition
```
in dummy app
did `pnpm i --frozen-lockfile` and `pnpm start`

abeforgit (changes requested):

addon/components/confidentiality-plugin/redact.gts
abeforgit: note: also happens when I checkout 47f6d0c, the commit before the master merge

abeforgit: and also with #448 merged :/

lagartoverde: I wrote to @piemonkey yesterday because I was having the same error, thought maybe was some local problem

elpoelma: I think it is because it needs to be a default export, not a named export for it to work.

So e.g.:
```js
const Redact: TemplateOnlyComponent<Sig> = <template>
  <Mark
    @icon={{NotVisibleIcon}}
    @title={{t 'confidentiality-plugin.redact'}}
    @mark='redacted'
    @controller={{@controller}}
  />
</template>;

export default Redact;
```

abeforgit: confirmed that fixes it

abeforgit: this is probably only a problem because it's used in a non-gts component

elpoelma: yep

PR #456, opened 2024-07-15
fix(snippets): make snippet nodes accept any block as content

abeforgit:
### Overview

The previous content spec crashed in a schema without the mentioned nodes, like a decision. It turns out you can only mention things in the content spec that exist in your schema.

In any case, we're moving more and more towards looser content specs as that works way better. We should strive to use `block+` in most block-style nodes. 

This does break compatibility with the "old" article-structures, as they are not part of the block group. But since the feature is built for the IV, this is fine.

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/thick-schools-nail.md
+++ b/.changeset/thick-schools-nail.md
@@ -1,0 +1,5 @@
+---
+'@lblod/ember-rdfa-editor-lblod-plugins': patch
+---
+
+Allow any block content in snippet nodes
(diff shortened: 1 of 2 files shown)


elpoelma (approved):
Looks good to me! In the future, we will probably replace those `title | chapter | article` structures by the new structure node anyway, which should be part of the 'block' group.

PR #463, opened 2024-08-13
GN-4949: person-variable-plugin: fix issue with unresponsive edit/insert button

elpoelma:
### Overview
This PR fixes the issue with the unresponsive edit/insert button when a person variable is selected.
Specifically, it replaces the `trackedFunction`  by a simple getter.

**Cause of the issue**
   -  The edit/insert card is only shown when selectedPersonNode is not loading and has resolved to a value
   - The selectedPersonNode trackedFunction is dependent on the editor state
   - When clicking on the edit/insert button outside the editor, a click event is fired => this causes the editor state to change (as the selection is no longer visible)
   - The selectedPersonNode trackedFunction is triggered due to a state change and goes into 'loading' mode, which means it no longer has a value (as it is pending) => this causes the card to hide (for only a few ms, but that's enough)
   - Due to the card being hidden for a few ms, the click event never reaches the button handler => the click handler is never triggered

[GN-4949](https://binnenland.atlassian.net/browse/GN-4949?atlOrigin=eyJpIjoiYzFiNTc5ZjgwODA3NDU0OWI1MDUwZjdmNmIyZDhmZTgiLCJwIjoiaiJ9)

### How to test/reproduce
- Start the dummy app
- Insert a person variable + click outside the variable
- Select the person variable again
- Notice that the edit/insert buttons are now directly responsive

### Challenges/uncertainties
I do not directly see a reason why to use a `trackedFunction` in this situation, especially as we aren't dealing with an async function/loading states. Correct me if I'm mistaken.

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/strange-fishes-lay.md
+++ b/.changeset/strange-fishes-lay.md
@@ -1,0 +1,5 @@
+---
+'@lblod/ember-rdfa-editor-lblod-plugins': minor
+---
+
+Fix issue with unresponsive edit/insert button when person variable is selected
--- a/addon/components/variable-plugin/codelist/edit.ts
+++ b/addon/components/variable-plugin/codelist/edit.ts
@@ -33,7 +33,7 @@ export default class CodelistEditComponent extends Component<Args> {
     return this.args.controller;
   }
 
-  selectedCodelist = trackedFunction(this, () => {
+  get selectedCodelist() {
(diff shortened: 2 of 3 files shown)


abeforgit (approved):
:100:

PR #465, opened 2024-08-15
Correctly export empty person nodes

lagartoverde:
### Overview
Added export to an empty person node to mimic what the user sees in the nodeview. To discuss if this should be like this or we should change both texts (or any of the two at least) to something like "Insert person".

GN-4978

### Setup
None

### How to test/reproduce
Go to the regulatory statements tab, insert a person variable and click on the show raw export button to see that the variable gets correclty exported

### Challenges/uncertainties
As I said maybe we should change the text

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/modern-cougars-compete.md
+++ b/.changeset/modern-cougars-compete.md
@@ -1,0 +1,5 @@
+---
+'@lblod/ember-rdfa-editor-lblod-plugins': patch
+---
+
+Correctly export empty person nodes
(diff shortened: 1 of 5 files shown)


elpoelma (commented):
Looks good,
I would change the placeholder to:
- nl: 'Voeg persoon in'
- en: 'Insert person'
That way, it's consistent with the new location, number and date variables.

PR #469, opened 2024-08-26
mandatee-table plugin: remove insertion constraint on mandatee table insertion

elpoelma:
### Overview
This PR removes the insertion constraint on mandatee tables. This means mandatee tables may be inserted everywhere (including snippets).

None

### How to test/reproduce
- Start the dummy app
- Notice that you can insert a mandatee table everywhere in the document (not just in a decision)
- Note: the synchronization of a mandatee table can only be done inside a decision (with the current config).

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/cold-schools-heal.md
+++ b/.changeset/cold-schools-heal.md
@@ -1,0 +1,5 @@
+---
+'@lblod/ember-rdfa-editor-lblod-plugins': patch
+---
+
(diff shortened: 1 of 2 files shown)


lagartoverde (approved):
Works perfectly fine maybe we should disable the button if you are not in a decision, or do something else rather than just throwing an error

PR #472, opened 2024-09-03
Gn 4901 autofill variable

lagartoverde:
### Overview
Made a new variable that get's autofilled if the correct context is provided

GN-4901

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations


lagartoverde:
For testing: https://github.com/lblod/frontend-gelinkt-notuleren/pull/715 and https://github.com/lblod/frontend-reglementaire-bijlage/pull/271

addon/components/variable-plugin/autofilled/insert.hbs
```diff
@@ -0,0 +1,32 @@
...
+    {{on 'input' this.updateAutofillKey}}
+  />
+</AuFormRow>
+<AuFormRow>
+  <AuCheckbox
+    id='convert_to_string'
+    @checked={{this.convertToString}}
+    @onChange={{this.updateConvertToString}}
+  >
+    {{t 'variable-plugin.autofill.convertToString'}}
```
piemonkey, line 27: When first looking at this, it wasn't clear to me what this meant. Maybe something that refers to leaving the variable node intact or not would be more clear? Maybe also an information text of some kind? As it is, it's not clear if, for example, this would be necessary to ensure that a date is inserted as a string rather than the timestamp.

addon/components/variable-plugin/autofilled/insert.hbs
```diff
@@ -0,0 +1,32 @@
...
+  <VariablePlugin::Utils::LabelInput
+    @label={{this.label}}
+    @updateLabel={{this.updateLabel}}
+  />
+</AuFormRow>
+<AuFormRow>
+  <AuLabel for='autofill_key'>
+    {{t 'variable-plugin.autofill.autofillKey'}}
+  </AuLabel>
+  <AuNativeInput
```
piemonkey, line 12: I think it would make more sense to me as a powerselect, with the possible values taken from configuration. otherwise there is too much margin for error.

addon/components/variable-plugin/autofilled/nodeview.hbs
```diff
@@ -0,0 +1,25 @@
+{{! @glint-nocheck: not typesafe yet }}
+<AuPill
+  @icon={{this.PencilIcon}}
+  @iconAlignment='right'
+  class='variable'
+  {{on 'click' this.onClick}}
+>
+  <EmberNode::EmbeddedEditor
```
piemonkey, line 8: Do we actually need an embedded editor? As it is you can actually replace the content with text, which does correctly get replaced with the value, but it still seems confusing to the user.

piemonkey (commented):
Works well but some comments on the UX.

Also, I think we need an edit component for this to be useful, as currently there's no easy way to modify it, you need to insert a new one.

lagartoverde:
Ok, so after rewriting a lot of the code, the variable will be initialized only once, so the user can modify it and it won't override its changes

addon/components/variable-plugin/autofilled/insert.hbs
lagartoverde: Changed the text, check if it's more clear or feel free to suggest one :)

addon/components/variable-plugin/autofilled/insert.hbs
lagartoverde: Discussed with Elena but will be happy to discuss with you in a standup or we can set up a meeting, the idea is that the plugin is as flexible as possible and the template app have no idea which final values will be implemented in the filler app, this of course has inconvenients as it has margin for error

addon/components/variable-plugin/autofilled/nodeview.hbs
lagartoverde: Fixed that, thank you for spotting the problem :)

piemonkey: It seems like you understood something different than I meant from my comment, but I'm glad it helped spot a bug.

What I meant was, that I don't see why we want an embedded editor here at all. To my mind, the contents of the variable should only be changeable by setting a different key or label. Here's a screenshot to demonstrate what I mean. This is a document with 2 autofill variables. You can see that I typed over the contents of the second, which means the only way to know that it's an autofill variable is to look in the sidebar. Is there some reason that template builders would want to be able to do this that I am missing?

image

addon/components/variable-plugin/autofilled/insert.hbs
piemonkey: Hmm, that makes sense but it still seems error-prone since we know at least some common ones that are likely to exist. Happy to just go with a text input though.

addon/components/variable-plugin/autofilled/edit.gts
```diff
@@ -0,0 +1,174 @@
...
+  get schema() {
+    return this.args.controller.schema;
+  }
+
+  get documentLanguage() {
+    return this.controller.documentLanguage;
+  }
+
+  @action
+  updateLabel(event: InputEvent) {
```
piemonkey, line 87: This isn't used, but I think we should have a way to edit the label too.

addon/components/snippet-plugin/snippet-insert.gts
```diff
@@ -75,9 +75,11 @@ export default class SnippetInsertComponent extends Component<Sig> {
         (tr, snippet) => tr.replaceSelectionWith(snippet),
```
piemonkey, line 75: Do we need to do a `tr.setMeta('insertSnippet', true);` somewhere in this branch?

lagartoverde: It's on the createAndInsertSnippet function

addon/components/variable-plugin/autofilled/edit.gts
lagartoverde: Done

piemonkey (approved):

translations/nl-BE.yaml
```diff
@@ -273,6 +275,10 @@ variable-plugin:
   person:
     card-title: Persoon invoegen
     nodeview-placeholder: Voeg persoon in
+  autofill:
+    autofillKey: Autovultoets
```
abeforgit, line 279:
```suggestion
    autofillKey: Sleutelwaarde
```

translations/nl-BE.yaml
```diff
@@ -273,6 +275,10 @@ variable-plugin:
...
+    convertToString: Variabele node verwijderen bij tekst invoegen
+    autofillKeyPlaceholder: Autovultoets
```
abeforgit, line 281:
```suggestion
    autofillKeyPlaceholder: Sleutelwaarde
```

translations/nl-BE.yaml
```diff
@@ -273,6 +275,10 @@ variable-plugin:
   person:
     card-title: Persoon invoegen
     nodeview-placeholder: Voeg persoon in
+  autofill:
+    autofillKey: Autovultoets
+    convertToString: Variabele node verwijderen bij tekst invoegen
```
abeforgit, line 280:
```suggestion
    convertToString: Vervang variabele door platte tekst bij invoegen
```

translations/en-US.yaml
```diff
@@ -269,6 +271,10 @@ variable-plugin:
   person:
     card-title: Insert person
     nodeview-placeholder: Insert person
+  autofill:
+    autofillKey: Autofill Key
+    convertToString: Remove variable node on text insertion
```
abeforgit, line 276:
```suggestion
    convertToString: Replace variable node with plain text on text insertion

abeforgit (changes requested):
some translation improvements

abeforgit:
@lagartoverde you can merge after you do the translations

PR #473, opened 2024-09-04
Add error handling to lmb-plugin

piemonkey:
### Overview
Discovered that this wasn't wired up when looking into QA feedback.

Related to problems on GN: https://github.com/lblod/app-gelinkt-notuleren/pull/192
And RB: https://github.com/lblod/frontend-reglementaire-bijlage/pull/269
From Jira issue: [binnenland.atlassian.net/browse/GN-5018](https://binnenland.atlassian.net/browse/GN-5018)

### Setup
Just running the test-app without setting a reasonable endpoint works well as a demonstration of a network error.

### How to test/reproduce
Insert a person variable and then insert a person to open the modal.

### Challenges/uncertainties
N/A

### Checks PR readiness
- [ ] changelog
- [ ] npm lint
- [ ] no new deprecations

### DIFF
--- a/addon/components/lmb-plugin/list.hbs
+++ b/addon/components/lmb-plugin/list.hbs
@@ -37,7 +37,7 @@
     </tr>
   </:header>
   <:body>
-    {{#unless (or @services.isRunning @error)}}
+    {{#unless @services.isRunning}}
       {{#if @services.value.totalCount}}
         {{#each @services.value.results as |row|}}
           <tr>
--- a/addon/components/lmb-plugin/search-modal.ts
+++ b/addon/components/lmb-plugin/search-modal.ts
@@ -42,10 +42,17 @@ export default class LmbPluginSearchModalComponent extends Component<Args> {
     return fetchMandatees({ endpoint: this.args.config.endpoint });
   });
 
+  // TODO Either make this a trackedFunction or do filtering on the query and correctly pass an
+  // AbortController
   search = restartableTask(async () => {
     // Can't do what I want, so if the user modifies the filter before resolving the query will run again
     if (!this.fetchData.lastComplete) {
-      await this.fetchData.perform();
+      try {
+        await this.fetchData.perform();
+      } catch (err) {
+        console.error('Got an error fetching LMB data', err);
+        this.error = err;
+      }
     }
 
     if (!this.fetchData.lastComplete?.value) return;
--- a/.changeset/yellow-tables-live.md
+++ b/.changeset/yellow-tables-live.md
@@ -1,0 +1,5 @@
+---
+'@lblod/ember-rdfa-editor-lblod-plugins': patch
+---
+
+Add error handling to lmb-plugin
--- a/addon/components/lmb-plugin/search-modal.hbs
+++ b/addon/components/lmb-plugin/search-modal.hbs
@@ -36,13 +36,16 @@
       </mc.sidebar>
       <mc.content @scroll={{true}}>
         <div class='worship-modal--list-container'>
-          <LmbPlugin::List
-            @error={{this.error}}
-            @services={{this.servicesResource}}
-            @sort={{this.sort}}
-            @setSort={{this.setSort}}
-            @insert={{@onInsert}}
-          />
+          {{#if this.error}}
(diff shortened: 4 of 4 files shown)


addon/components/lmb-plugin/search-modal.ts
```diff
@@ -39,13 +39,24 @@ export default class LmbPluginSearchModalComponent extends Component<Args> {
...
 
   search = restartableTask(async () => {
     // Can't do what I want, so if the user modifies the filter before resolving the query will run again
     if (!this.fetchData.lastComplete) {
-      await this.fetchData.perform();
+      try {
+        await this.fetchData.perform();
+      } catch (err) {
+        console.error('got an error fetching data', err);
+        this.error = err;
```
elpoelma, line 58: This doesn't seem necessary to me? As we already are setting the `error` field in the `fetchData` task?

addon/components/lmb-plugin/list.hbs
```diff
@@ -37,6 +37,9 @@
     </tr>
   </:header>
   <:body>
+    {{#if @error}}
+      <Common::Search::AlertLoadError @error={{@error}} />
+    {{/if}}
```
elpoelma, line 42: I think it might be better to not show the table at all when there is an error? We could simply show the error without the table headers being shown.

addon/components/lmb-plugin/search-modal.ts
```diff
@@ -39,13 +39,24 @@ export default class LmbPluginSearchModalComponent extends Component<Args> {
   }
 
   fetchData = restartableTask(async () => {
```
elpoelma, line 41: Not really related to this PR: I don't really understand why this is a `restartableTask`. The `this.error` field assignment also doesn't really seem necessary to me, as the `lastComplete`/`last` already has an `error` field.
It works, but ideally I think a `trackedFunction` might be more suited in this situation. Ideally, I also think that the `search` function shouldn't really be a `task` instance at all, as it doesn't really run any async operations (except for this single `fetchData` run), but not sure about that.

elpoelma (changes requested):
The error message correctly shows up when there are connection problems.
I do have some small remarks. One of the remarks is not really related to this PR, but added it anyway.

elpoelma:
Also wondering how well this plugin will scale when the production data will contain a lot more people (as we shouldn't only fetch mandatees here, but rather all elected officials). It might be better to do paginated + filtered queries.

addon/components/lmb-plugin/search-modal.ts
piemonkey: Actually, weirdly, the other catch doesn't run. There must be something weird about how TS or ember is transpiling the code as if I `await` the `fetchMandatees()` call into a variable and then return it, the other catch runs as expected. I threw this in as a quick fix, I'll make a more carefully thought out version.

addon/components/lmb-plugin/list.hbs
piemonkey: I did this as I was just mimicking the existing API, which passed the error to the list component, even though it was never used. We could also just not show the List at all on error.

addon/components/lmb-plugin/search-modal.ts
piemonkey: I think it's a `restartableTask`  as changing the search term will re-run the search. We're not using an `AbortController` though, so I don't think it's adding anything right now over being a `trackedFunction`.

piemonkey: Actually, just looked again and realised that the search is just filtering on the results, so indeed it should not be restartable.

piemonkey:
I've updated this PR based on the feedback. I didn't change to a trackedFunction, as I thought it would be better to move to filtering on the search, and didn't want to include that extra work in this PR or waste time refactoring working code that will be moved back to a `restartableTask` if we do that filtering.

addon/components/lmb-plugin/search-modal.ts
elpoelma: Huh ok!

elpoelma (approved):

PR #474, opened 2024-09-05
LMB-plugin: query mandatees using `form-urlencoded` approach

elpoelma:
### Overview
This PR modifies the approach taken to query the mandatees in the LMB-plugin. It now makes use of the `executeQuery` utility function, which sends a SPARQL query in the `application/x-www-form-urlencoded` format. This also ensures that the service is compatible with any SPARQL endpoint as the plugin now uses a standardized way to send the SPARQL query.

None

### Setup
You can test this PR with both the current version of the `vendor-proxy` service (0.1.0) or with https://github.com/lblod/vendor-proxy-service/pull/3

You should have a GN stack running containing the `vendor-proxy` service.

### How to test/reproduce
The plugin should work as before:
- Open the LMB-mandatee modal
- The correct mandatees should load

### Challenges/uncertainties
Unsure why a custom `json`-based approach was taken to send SPARQL queries. LMK if there is something I missed about disadvantages/advantages of using the `urlencoded` approach.

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/khaki-ligers-march.md
+++ b/.changeset/khaki-ligers-march.md
@@ -1,0 +1,5 @@
+---
+'@lblod/ember-rdfa-editor-lblod-plugins': patch
+---
+
+LMB-plugin: query mandatees using `application/x-www-form-urlencoded` approach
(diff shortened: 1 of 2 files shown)


piemonkey (approved):
Looks good for me when tested linked to GN.

PR #475, opened 2024-09-05
mandatee-table plugin: replace title `input` by `textarea`

elpoelma:
### Overview
This PR replaces the `title` `input` element by a `textarea` element. This ensures that it is easier for template creators to input long titles.

None

### How to test/reproduce
- Start the dummy application
- Insert a mandatee table
- Notice that you can now edit the table title using a `textarea` element

### Checks PR readiness
- [ ] changelog
- [ ] npm lint
- [ ] no new deprecations

### DIFF
--- a/.changeset/lazy-bananas-add.md
+++ b/.changeset/lazy-bananas-add.md
@@ -1,0 +1,5 @@
+---
+'@lblod/ember-rdfa-editor-lblod-plugins': minor
+---
+
+mandatee-table plugin: replace `input` element by `textarea` element for entering mandatee table titles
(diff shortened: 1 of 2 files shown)


piemonkey:
Works well, the only thing I'm not sure about is newlines. The textarea allows newlines, but they don't show in the title. I guess this is fine though as I can't think of a good way to handle it otherwise.

elpoelma:
> Works well, the only thing I'm not sure about is newlines. The textarea allows newlines, but they don't show in the title. I guess this is fine though as I can't think of a good way to handle it otherwise.

True, I guess I could also replace the `p` element with `strong` inside by an `h` element which does allow newlines?

elpoelma:
> Works well, the only thing I'm not sure about is newlines. The textarea allows newlines, but they don't show in the title. I guess this is fine though as I can't think of a good way to handle it otherwise.

We could also add `whitespace: pre-line` to the heading.

elpoelma:
@piemonkey I added the following:
- Replaced the `p` + `strong` combo by an `h6` (which should be more semantically correct)
- Added `white-space: pre-line` to `h6` element to support newlines.

piemonkey (approved):

lagartoverde (approved):
Works fine, maybe it's good to expand the text area and the card if it has more lines, but like that is already good

elpoelma:
@lagartoverde I added a custom `auto-resizing-text-area` component which should work in both browsers.

PR #476, opened 2024-09-05
Add content string configuration option to snippet plugin

piemonkey:
### Overview
~~Brings snippet contents more in line with the content that can be created in RB and export a constant to be used there to ensure only valid snippet contents are created.~~
Allows snippet node allowed content string to be configured when inserting into the schema. This means that we can only include node types that the schema actually supports. When done correctly, this means that there is no snippet that can fail to insert due to using nodes that do not exist, but it does mean that snippets can still appear strangely as they use a nodespec that doesn't exist in the schema, therefore prosemirror just tries it's best with what it has.

Part of https://binnenland.atlassian.net/browse/GN-5019 but not the complete fix.

### Setup
You'll need to get snippets from somewhere you can create ones with nodes which aren't in the `block` group, such as document_title, e.g. a local RB stack.

### How to test/reproduce
Insert a snippet placeholder with the relevant snippet list and then insert the snippet which includes a node not in the schema. If this is in the content string but not in the schema, you will see an error, but if it either is in both, or is in neither, the snippet should insert correctly (though look different in each case).

### Challenges/uncertainties
N/A

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations


elpoelma (commented):
Not really sure if this extra API is necessary, as you should be able to override the `content` setting when adding the node to your schema in the following matter:
```
snippet: {
  ...snippet,
  content: ...,
}
```
I do think a general API to reconfigure existing node-spec would be nice though, e.g. 
`snippet.configure(...)`, which would return a new node-spec.

PR #478, opened 2024-09-08
Recalculate article numbers after inserting snippet

abeforgit:
### Overview

BREAKING: needs latest editor for the headless changes so the command doesn't need to depend on the view, only breaking because of peerdep bump

Does what it says on the tin, runs the recalc function when snippets are inserted/deleted to make any potential articles line up

also a slight refactor of the insertion logic that was duplicated in 2 locations into a command
### Setup

Developed while yalced to GN, don't know if testable in dummy app
### Checks PR readiness

- [x] changelog
- [x] npm lint
- [x] no new deprecations


piemonkey (approved):
This all looks good to me. I had to do some significant merge conflict resolution though, so it's maybe best if someone else looks at this too. The conflicts were with the imported resources work as both changed how snippets are inserted.

PR #480, opened 2024-09-17
Update ember-resources

piemonkey:
### Overview
Based on [PR from NullVoxPopuli](https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/477), but with master merged, a changeset and minor tweaks.

The original: https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/477

### Setup
N/A

### How to test/reproduce
Anything using ember-resources works, e.g. search boxes. Also should work when linked to GN or RB frontends.

### Challenges/uncertainties
This will require an upgrade from consumer apps that use ember dependency-lint, but according to testing, the versions can actually co-exist peacefully. I'm unsure if we should therefore merge this or whether we should wait until we do a major release.

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations


elpoelma (commented):
Works well :smile: 
I think we can at least move to the `reactiveweb` package without issues. Not so sure about the `ember-resources` package, as we kind of still need to take the app-folder merging into account. I would assume that in this case it uses the host-app version of ember-resources, but I'm not completely sure. Maybe (if we do not want to break the API), we should remain on `ember-resources` 6.5.1 for now, but already move to `reactiveweb`.

Also w

piemonkey:
@elpoelma I've tested both (my original version with the ember-resources upgrade and the same with ember-resources not upgraded) and both seem to work fine included in GN as dev versions from npm (both fail with yalc). The only failure I see is ember-dependency-lint, which is being very paranoid. Personally, I think since we control the consuming apps, we can safely upgrade, but I'll change this for the version with ER 6.5 so we can avoid messing with any other consumers.

piemonkey:
Actually, I just realised that https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/478 will mean a major release due to dependency changes, so should we remove my latest commit (so we use ER 7.0) and just have an extra dependency change required for that version?

elpoelma:
> Actually, I just realised that #478 will mean a major release due to dependency changes, so should we remove my latest commit (so we use ER 7.0) and just have an extra dependency change required for that version?

I also have a PR with breaking changes coming up (the snippet rework) so we can add that one too.

PR #485, opened 2024-10-07
Move back to removing snippet buttons instead of making them invisible

piemonkey:
### Overview
It seems that we don't actually need to make the buttons invisible if we have `contenteditable="true"`. It looks like the issue is that prosemirror was trying to put the selection into the help text for the buttons. This removes a weird hacky almost-fix that is best not to leave in the codebase to surprise us in the future.

No changeset as it's a purely internal change.

Undoes a hacky part of https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/479

### Setup
N/A

### How to test/reproduce
Needs to be tested in Chrome, with snippets which fill the width of the editor multiple times, e.g. a big block of text.

### Challenges/uncertainties
N/A

### Checks PR readiness
- [ ] changelog
- [ ] npm lint
- [ ] no new deprecations


abeforgit (approved):
lgtm

PR #487, opened 2024-10-15
IRGN - add traffic sign prefix

andreo141:
### Overview
Governments want to be able to see the type of traffic sign prefixed in the generated rdfa when adding a traffic measure.

https://binnenland.atlassian.net/browse/GN-4882

### How to test/reproduce
Try to add a traffic sign. Depending on the type of sign the prefix should change.

### Challenges/uncertainties
Not super familiar with IRGN and the editor, so all feedback is welcome. Based on the ticket description, this is what i understood that needed to happen.

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations


lagartoverde (changes requested):

andreo141:
I modified the query where the signs are being fetched and included the signtype in the SELECT portion of the query.

addon/models/sign.ts
```diff
@@ -4,6 +4,7 @@ import { unwrap } from '../utils/option';
 export default class Sign {
   constructor(
     readonly code: string,
+    readonly signType: string,
```
lagartoverde, line 7: You can rename this type as it's already inside the sign :)

lagartoverde (approved):
Works for me, weird that we don't add this type as rdfa info but indeed the ticket doesn't ask for that, maybe @nvdk can throw some light?

addon/models/sign.ts
nvdk: that's ok for now

andreo141:
This can be merged but it seems like pipeline is still waiting for approval

PR #488, opened 2024-10-16
Reset person edit modal on close

lagartoverde:
### Overview
Reset the person edit modal so it doesn't show old filtered data when it's open again

GN-5096

### Setup
You must have a working vendor proxy service in the backend, or you can test linking the plugins to GN-frontend and proxying that to dev, let me know if you need any help

### How to test/reproduce
Insert 2 person variable nodes in the RB part, then swich to the GN part and try to edit them, you should be able to filter on one, close the modal and open the same one or the other and the filter will be cleared and the data refreshed

### Challenges/uncertainties
None

### Checks PR readiness
- [ ] changelog
- [ ] npm lint
- [ ] no new deprecations

### DIFF
--- a/addon/components/lmb-plugin/search-modal.ts
+++ b/addon/components/lmb-plugin/search-modal.ts
@@ -34,6 +34,8 @@ export default class LmbPluginSearchModalComponent extends Component<Args> {
 
   @action
   async closeModal() {
+    this.inputSearchText = null;
+    this.sort = false;
     await this.servicesResource.cancel();
     this.args.closeModal();
   }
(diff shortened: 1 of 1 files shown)


abeforgit (changes requested):
only resets the search the second time modal is opened
so:
- open modal
- enter search, table updates
- close modal, select a different place in the doc
- insert new mandatee
- observe table is still filtered
- close modal
- open again: now table is reset

lagartoverde:
> only resets the search the second time modal is opened so:
> 
> * open modal
> * enter search, table updates
> * insert mandatee
> * select a different place in the doc
> * open modal again
> * observe table is still filtered
> * close modal without inserting (if you insert here, the modal still doesn't reset)
> * open again: now table is reset

Yes my bad, I was resetting every time the modal is close, but the insert was closing the modal without using the function, I solved that

PR #493, opened 2024-10-31
Redesign structure nodes v1

lagartoverde:
### Overview
Added some css to match Saska design https://www.figma.com/design/tfBSv7999dzz1sy6svMfsY/GN_editor?node-id=81-575&node-type=frame&t=GxqZjDXLUvDzepry-0

GN-5118

### Setup
None

### How to test/reproduce
Insert an article and see the design, you can also inserte several articles one inside another 

### Challenges/uncertainties
We will have to recheck the padding on the left when adding the buttons

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/eleven-months-know.md
+++ b/.changeset/eleven-months-know.md
@@ -1,5 +1,0 @@
----
-'@lblod/ember-rdfa-editor-lblod-plugins': patch
----
-
-Do not add `" ."` at the end of 'only article' for single articles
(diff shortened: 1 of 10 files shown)


abeforgit (changes requested):
missing some border radius:
image


also the selected state doesn't match the design (only the outer border is blue, should be like in the above screenshot)

lagartoverde:
> missing some border radius: image
> 
> also the selected state doesn't match the design (only the outer border is blue, should be like in the above screenshot)

Talked with saska, without the border radius in the bottom is better, and did the active status now

PR #494, opened 2024-11-04
GN-5153: only trigger event handlers of snippet actions once

elpoelma:
### Overview
This PR solves the issue of the snippet buttons' event handlers being fired twice (thus why two snippets were removed at once).
It seems that there were some leftovers of the attempt to resolve the chromium button bug: the button event handlers were tied to both the button and the icon inside it, which results in the handler being fired twice.

[GN-5153](https://binnenland.atlassian.net/browse/GN-5153)

### Setup
None

### How to test/reproduce
**Before**
- Start the dummy app
- Insert two snippets after eachother
- Click on the icon inside the remove button on the first snippet
- Notice that both snippets are removed (due to the event handler being fired twice)

**After**
- Insert two snippets after eachother
- Click on the icon inside the remove button on the first snippet
- Only the relevant snippet should be removed (and the event handler is only fired once)

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/fuzzy-pillows-relax.md
+++ b/.changeset/fuzzy-pillows-relax.md
@@ -1,0 +1,5 @@
+---
+'@lblod/ember-rdfa-editor-lblod-plugins': patch
+---
+
+Ensure that the event handlers of snippet actions are only triggered once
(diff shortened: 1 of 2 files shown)


abeforgit (approved):
good catch :+1:

PR #495, opened 2024-11-05
Replace snippet placeholder when deleting last snippet

piemonkey:
### Overview
Tried to reduce some of the differences in naming and arguments between snippet nodes and snippet placeholders while doing this, so it's easier to do other work that pulls them together in the future but without re-writing everything. Also fixes commas in snippet list names.

Should be backwards compatible but does (can) not fix some historical issues that will remain in old documents. Namely:
- Snippets inserted in old documents don't know which placeholder they 'came from', so will be handled as separate 'groups'. For example, if you created a document from a template including a placeholder, inserted two snippets, then saved; the new code will now, when deleting either of those snippets, replace it with a placeholder. This can easily be deleted, by pressing backspace, so this shouldn't be too much of a problem.
- Snippet placeholders which include a snippet with a `,` in the name will remain broken. There's no reasonable way for us to figure out that this has happened, so we just handle it as before.

Jira ticket: https://binnenland.atlassian.net/browse/GN-5003

### Setup
N/A

### How to test/reproduce
Create a document with placeholders, multi-snippet placeholders and placeholders with commas in the list names and play around inserting and deleting snippets and refreshing. 

### Challenges/uncertainties
N/A

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/gentle-berries-exist.md
+++ b/.changeset/gentle-berries-exist.md
@@ -1,0 +1,5 @@
+---
+'@lblod/ember-rdfa-editor-lblod-plugins': minor
+---
+
+When removing last snippet of a 'group', replace the placeholder instead of completely deleting
--- a/.changeset/hungry-trees-perform.md
+++ b/.changeset/hungry-trees-perform.md
@@ -1,0 +1,5 @@
+---
+'@lblod/ember-rdfa-editor-lblod-plugins': patch
+---
+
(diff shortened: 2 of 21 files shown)


abeforgit:
completely fair compromises wrt backwards compat, will have a look tomorrow

elpoelma (approved):
Looks good to me! Quite intuitive :)

PR #496, opened 2024-11-05
Snippet placeholder redesign

piemonkey:
### Overview
Continued a little of the refactoring from replacing the snippet placeholder, so that the snippet placeholder component is now used inside a snippet component. This means that I could re-use the insert logic from the snippet buttons and if we wanted in the future we could include buttons in the same style as for inserted snippets. Since I was not including buttons for now, I was able to use @elpoelma 's redesign implementation work almost untouched.

For consistency, I move the snippet list ids information to RDFa for inserted snippet nodes.

Builds on https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/495
Jira ticket: https://binnenland.atlassian.net/browse/GN-5117

### Setup
N/A

### How to test/reproduce
Insert, replace, delete snippets and snippet placeholders with different configurations and make sure everything works as expected. The expected interactivity is different depending on which sample page:
- Besluit: It's only possible to insert snippets using the in-node insert button. placeholders cannot be inserted
- Reg statement: The in-node insert snippet button is hidden, but the sidebar button remains.
- Snippet edit: The in-node snippet insert button is shown, the sidebar button is not there. Placeholders can be inserted.

### Challenges/uncertainties
N/A

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/app/styles/snippet-plugin.scss
+++ b/app/styles/snippet-plugin.scss
@@ -145,14 +145,14 @@
 .say-snippet-placeholder {
   display: flex;
   flex-direction: row;
-  padding: $au-unit-small $au-unit-small*1.5;
+  padding: $au-unit-small $au-unit-small * 1.5;
   background-color: #fff7bb;
   border-radius: var(--au-radius);
   border: 0.1rem solid #e5d34f;
 }
 
 .say-snippet-placeholder__icon {
-  background-color: #FFA10A;
+  background-color: #ffa10a;
   color: black;
   height: $au-unit - 0.1rem; // compensate for visual distortion of perfect circle
   width: $au-unit;
@@ -169,7 +169,7 @@
 }
 
 .say-snippet-placeholder__title {
-  color: #9E5805;
+  color: #9e5805;
   font-weight: 500;
 }
 
--- a/app/styles/snippet-plugin.scss
+++ b/app/styles/snippet-plugin.scss
@@ -146,9 +146,9 @@
   display: flex;
   flex-direction: row;
   padding: $au-unit-small $au-unit-small * 1.5;
-  background-color: #fff7bb;
+  background-color: var(--au-orange-300);
   border-radius: var(--au-radius);
-  border: 0.1rem solid #e5d34f;
+  border: 0.1rem solid var(--au-orange-500);
 }
 
 .say-snippet-placeholder__icon {
@@ -169,7 +169,7 @@
 }
 
 .say-snippet-placeholder__title {
-  color: #9e5805;
+  color: var(--au-orange-700);
   font-weight: var(--au-medium);
 }
 
--- a/app/styles/snippet-plugin.scss
+++ b/app/styles/snippet-plugin.scss
@@ -152,7 +152,7 @@
 }
 
 .say-snippet-placeholder__icon {
-  background-color: #ffa10a;
+  background-color: var(--au-orange-500);
   color: black;
   height: $au-unit - 0.1rem; // compensate for visual distortion of perfect circle
   width: $au-unit;
@@ -170,7 +170,7 @@
 
 .say-snippet-placeholder__title {
   color: #9e5805;
-  font-weight: 500;
+  font-weight: var(--au-medium);
 }
 
 .say-snippet-placeholder__button {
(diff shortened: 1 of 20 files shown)


app/styles/snippet-plugin.scss
```diff
@@ -164,3 +141,62 @@ div[typeof='besluitpublicatie:Documentonderdeel']
     }
   }
 }
+
+.say-snippet-placeholder {
+  display: flex;
+  flex-direction: row;
+  padding: $au-unit-small $au-unit-small * 1.5;
+  background-color: #fff7bb;
```
lagartoverde, line 149: Aren't these colors already a variable? If not talk with saska as it should be a mistake

app/styles/snippet-plugin.scss
```diff
@@ -164,3 +141,62 @@ div[typeof='besluitpublicatie:Documentonderdeel']
...
+  flex-shrink: 0;
+}
+
+.say-snippet-placeholder__content {
+  margin-top: 0 !important;
+}
+
+.say-snippet-placeholder__title {
+  color: #9e5805;
+  font-weight: 500;
```
lagartoverde, line 173: var(--au-medium) or something like that

app/styles/snippet-plugin.scss
```diff
@@ -164,3 +141,62 @@ div[typeof='besluitpublicatie:Documentonderdeel']
...
+  margin-right: $au-unit * 0.5;
+  border-radius: $au-unit-large;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  flex-shrink: 0;
+}
+
+.say-snippet-placeholder__content {
+  margin-top: 0 !important;
```
lagartoverde, line 168: Is it possible not to add important here? maybe with more specifity, if not is okay

tests/dummy/app/controllers/snippet-edit-sample.ts
```diff
@@ -290,7 +288,7 @@ export default class RegulatoryStatementSampleController extends Controller {
           'https://publicatie.gelinkt-notuleren.vlaanderen.be/id/adres/',
       },
       snippet: {
-        endpoint: 'https://dev.reglementairebijlagen.lblod.info/sparql',
+        endpoint: 'https://dev.reglementairebijlagen.lblod.info/raw-sparql',
```
lagartoverde, line 291: why are we changing this to the uncached endpoint?

lagartoverde (changes requested):
Left some minor comments

tests/dummy/app/controllers/snippet-edit-sample.ts
elpoelma: Probably a mistake on my part :grimacing:

piemonkey: Oh, no. It was me. I spotted that the other two sample pages used the uncached one, so I made this the same. It's true we should really go the other way and make them all cached.

piemonkey:
@elpoelma are you looking at these comments or should I?

elpoelma:
> @elpoelma are you looking at these comments or should I?

You can do it if you want :) I am currently working on an editor PR which fixes the specificity css styling regression which should remove the need of the `!important` modifier.

app/styles/snippet-plugin.scss
piemonkey: I've checked with Saska but most of these colours aren't in AU.

app/styles/snippet-plugin.scss
piemonkey: Waiting on @elpoelma's fix: https://github.com/lblod/ember-rdfa-editor/pull/1226

app/styles/snippet-plugin.scss
piemonkey: I've heard back now and the intent was to use some different AU colours, so I've changed those now.

app/styles/snippet-plugin.scss
elpoelma: Was able to fix it, it was also still another issue in the end.

lagartoverde (approved):
Looks good to me

PR #503, opened 2024-11-08
GN-5209: extend/generalize 'nodes' config option to citation plugin

elpoelma:
### Overview
This PR extends/generalizes the 'nodes' config option of the citation plugin.
It now allows you to specify a condition which an active/context node needs to satisfy.
```ts
interface CitationPluginNodeConditionConfig {
  type: 'nodes';
  regex?: RegExp;

  activeInNode(node: PNode, state?: EditorState): boolean;
}
```
The previously expected `activeInNodeTypes` option is marked as deprecated and will be removed in a future release.
Currently, we were still using the  `activeInNodeTypes` option in the dummy app and GN. This failed however, as we do not use the 'motivering' node-spec/type anymore.

[GN-5209](https://binnenland.atlassian.net/browse/GN-5209)
https://github.com/lblod/frontend-gelinkt-notuleren/pull/763

### Setup
None

### How to test/reproduce
- Start the dummy app
- Open the decision sample page
- Insert a decision
- Ensure that you can insert a citation in the 'motivering' block.

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/addon/components/citation-plugin/citation-insert.ts
+++ b/addon/components/citation-plugin/citation-insert.ts
@@ -3,6 +3,7 @@ import { action } from '@ember/object';
 import { tracked } from '@glimmer/tracking';
 import {
   Fragment,
+  PNode,
   SayController,
   Slice,
   Transaction,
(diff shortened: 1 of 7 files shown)


addon/components/citation-plugin/citation-insert.ts
```diff
@@ -62,28 +63,32 @@ export default class EditorPluginsCitationInsertComponent extends Component<Args
...
+      let condition: (node: PNode) => boolean;
+      if ('activeInNodeTypes' in config) {
+        const nodeTypes = config.activeInNodeTypes(
+          this.controller.schema,
+          this.controller.mainEditorState,
+        );
+        condition = (node) => nodeTypes.has(node.type);
+      } else {
+        condition = (node) =>
+          config.activeInNode(node, this.controller.mainEditorState);
```
abeforgit, line 85: We probably typed it this way, but it would probably be best to handle the absence of any config, regardless of the config type, by enabling in the whole doc

same for the absence of the 'type' config

abeforgit (changes requested):

addon/components/citation-plugin/citation-insert.ts
elpoelma: Makes sense indeed, I'll adjust it.

elpoelma: is adjusted :)

PR #504, opened 2024-11-14
GN-5217: person variable adjustments

elpoelma:
### Overview
This PR contains 2 breaking changes:

**Adjustments to `person` variable:**
- Removal of unused `content` attribute
- Removal of `mandatee` attribute. Replaced by `value` attribute which contains an object of type `Person`.
  ```ts
  type Person = {
    uri: string;
    firstName: string;
    lastName: string;
  };
  ```
- Adjust parsing rules of `person_variable` nodespec to work with old and new serializations
- Is now more generic, instead of relying on `Mandatee` instances

**Adjustments to `lmb-plugin`:**
- Rework `fetchMandatees` queries into `fetchElectees`
  * Fetches all electees for a certain legislation period + all non-elected people with a mandate
  * Replace `Mandatee` model by simpler `Electee` model
- Adjust UI components to reflect the `Mandatee` to `Electee` change

TODOs:
- [x] Show a message in the `Kieslijst` column if the person is not an electee.
- [x] Show a message in the table if no results were found

[GN-5217](https://binnenland.atlassian.net/browse/GN-5217)
Soft-depends on https://github.com/lblod/app-gelinkt-notuleren/pull/212
https://github.com/lblod/frontend-gelinkt-notuleren/pull/769

### Setup
- Correctly configure the `lmb` plugin, adjust the endpoint used if needed
- (Optional, when testing with non-elected people) Use https://github.com/lblod/app-gelinkt-notuleren/pull/212 as a backend.

### How to test/reproduce
- Start the dummy app
- Open the 'Besluit Sample' page
- In the right-side menu, insert an 'Electee'
- Ensure that in the modal, you can search for any electee/non-elected mandatee for any period
- You should be able to easily filter on the name of the person
- Sorting the columns should work as expected
- Inserting a person should work as expected
- Check the RDFa of the inserted person node
- The node should contain a `value` attribute with info on the person
- Reloading the page should keep the node intact

**Backwards compatibility**
- The `person` node-spec is backwards compatible with previous versions containing the `mandatee` attribute
- Copy a 'mandatee' variable from GN
- Paste it into the dummy app, it should be parsed correctly

### Challenges/uncertainties
- Maybe it should be better to rename the `lmb-plugin` into `electee-plugin`
- I tried to make the person variable as generic as possible, so it is not dependent on electee or mandatee concepts

### Checks PR readiness
- [x] changelog
- [x] npm lint


addon/plugins/variable-plugin/variables/person.ts
```diff
@@ -58,16 +78,16 @@ const parseDOM = [
 
 const serialize = (node: PNode, state: EditorState): DOMOutputSpec => {
   const t = getTranslationFunction(state);
-  const mandatee = node.attrs.mandatee as Mandatee;
+  const person = node.attrs.value as Person | undefined;
   return renderRdfaAware({
     renderable: node,
     tag: 'span',
     attrs: {
       ...node.attrs,
```
abeforgit, line 86: this renders out the bare "value" attribute, with a value of "[object Object]", which is a bit ugly

abeforgit (changes requested):
The main problem I see is that the person's information is no longer visible in rdfa at all 

Maybe this was never the case? but at least the person URI should be retrievable from an rdfa parser, and preferable also the names

addon/plugins/variable-plugin/variables/person.ts
elpoelma: Fixed that.

elpoelma:
> The main problem I see is that the person's information is no longer visible in rdfa at all
> 
> Maybe this was never the case? but at least the person URI should be retrievable from an rdfa parser, and preferable also the names

This was indeed never the case, but I can look into annotating both the person URI and names.

abeforgit:
Here's a ticket for annotating the uri and names: https://binnenland.atlassian.net/browse/GN-5240 
this pr is good to go as-is

PR #505, opened 2024-11-14
Use the new template uris on variables

lagartoverde:
### Overview
When we use the variables on the final agendapoint the uris are the same, so if we insert 2 snippets with the same variable the rdfa parser does weird things with the values

GN-5232

### Setup
None

### How to test/reproduce
create a snippet in RB with variables, insert that snippet 2 times in GN, observe that the variables have different uuids on GN even if they come from the same snippet

### Challenges/uncertainties
None

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations


elpoelma:
@lagartoverde Shouldn't we make this configurable/optional, as you don't want to use the template URIs when inserting in GN directly I would think (e.g. for addresses and dates)

abeforgit:
> @lagartoverde Shouldn't we make this configurable/optional, as you don't want to use the template URIs when inserting in GN directly I would think (e.g. for addresses and dates)

yeah that's a good idea

addon/plugins/variable-plugin/recreateUuidsOnPaste.ts
```diff
@@ -72,12 +72,21 @@ function recreateUuidsOnNode(node: Node, schema: Schema) {
   const attrs = { ...node.attrs };
 
   attrs.properties = (attrs.properties as OutgoingTriple[]).map((prop) => {
+    console.log(prop);
```
elpoelma, line 75: console.log

addon/components/variable-plugin/address/utils.ts
elpoelma, line 1: We might also need to generate a placeholder for the `__rdfaId`, not sure though...

addon/components/variable-plugin/autofilled/insert.gts
elpoelma, line 1: We might need to generate a template URI for the `__rdfaId` here.

elpoelma (changes requested):
Looks good overall! Some minor comments...

addon/components/variable-plugin/address/utils.ts
elpoelma: oh ok!

elpoelma (approved):

elpoelma:
One thing: maybe we should also add a `templateMode` to https://github.com/lblod/ember-rdfa-editor-lblod-plugins/blob/1e760301a2e9ed161bc6e19a598bb1894a3acf2d/addon/components/lmb-plugin/insert.ts

PR #506, opened 2024-11-15
recreate mappings uri

lagartoverde:
### Overview
When recreating the variable uris we weren't recreating the mapping, which can result in unwanted behaviour

Maybe GN-5232

### Setup
None

### How to test/reproduce
Copy and paste a variable notice that the mapping uri also gets recreated, both if the uri is a template uri or a real uri

### Challenges/uncertainties
I'm looking at this pluggin and it's weird because I thought it was more general, than just recreating variable uris

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations


lagartoverde:
No needed anymore with https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/511

PR #510, opened 2024-11-20
Make decision-topic work outside decisions

abeforgit:
### Overview

Uses the new external triples feature to make the decision topic plugin operate
outside of a decision

adds a `decisionUri` config to the plugin, if present, the plugin will instead
insert external triples into the doc node

[GN-5170](https://binnenland.atlassian.net/browse/GN-5170)
[editor pr](https://github.com/lblod/ember-rdfa-editor/pull/1229)
[embeddable pr](https://github.com/lblod/frontend-embeddable-notule-editor/pull/269)
### Setup

editor dev build is already setup, so just install

### Challenges/uncertainties

The state logic of the plugin is ...odd... but I don't have time to rewrite it
to be more idiomatic so I mainly kept it the same

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations


piemonkey (commented):
Looks good. There is one weird behaviour, which I'm not sure is a problem or not. If you add any external triple with predicate `http://data.europa.eu/eli/ontology#is_about` and value anything but `https://data.vlaanderen.be/id/concept/BesluitThema/...` then it gets stripped as soon as the topic plugin adds or removes a topic. I don't think this will affect anything we do right now, but in theory it could.

abeforgit:
> Looks good. There is one weird behaviour, which I'm not sure is a problem or not. If you add any external triple with predicate `http://data.europa.eu/eli/ontology#is_about` and value anything but `https://data.vlaanderen.be/id/concept/BesluitThema/...` then it gets stripped as soon as the topic plugin adds or removes a topic. I don't think this will affect anything we do right now, but in theory it could.

great catch, it's even worse: the plugin wipes all externaltriples as soon as you set any topic

piemonkey (approved):
Looks good to me now.

PR #511, opened 2024-11-20
Feat/change recreate uris

lagartoverde:
### Overview
Implement new way of recreating uuuids on paste

### Setup
None

### How to test/reproduce
Create a few variables notice how the uuids of both the variable and the mapping get correctly recreated on paste

### Challenges/uncertainties
None

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/addon/plugins/variable-plugin/variables/date.ts
+++ b/addon/plugins/variable-plugin/variables/date.ts
@@ -36,6 +36,7 @@ import {
   RdfaAttrs,
   renderRdfaAware,
 } from '@lblod/ember-rdfa-editor/core/schema';
+import { recreateVariableUris } from '../utils/recreate-variable-uris';
 
 const TRANSLATION_FALLBACKS = {
   insertDate: 'Datum invoegen',
@@ -259,6 +260,7 @@ const emberNodeConfig = (options: DateOptions): EmberNodeConfig => ({
   selectable: true,
   draggable: false,
   atom: true,
+  recreateUriFunction: recreateVariableUris,
   defining: false,
   options,
   attrs: {


lagartoverde:
Do not merge before editor release

addon/plugins/variable-plugin/utils/recreate-variable-uris.ts
```diff
@@ -0,0 +1,43 @@
...
+    )
+  ) {
+    attrs.subject = `http://data.lblod.info/mappings/--ref-uuid4-${uuidv4()}`;
+  } else if (
+    attrs.subject &&
+    (attrs.subject as string).includes('http://data.lblod.info/mappings/')
+  ) {
+    attrs.subject = `http://data.lblod.info/mappings/${uuidv4()}`;
+  }
+  attrs.properties = (attrs.properties as OutgoingTriple[]).map((prop) => {
```
abeforgit, line 22: check out getOutgoingTriple and related utils in `utils/namespace.ts`

addon/plugins/variable-plugin/utils/recreate-variable-uris.ts
```diff
@@ -0,0 +1,43 @@
...
+    if (prop.predicate === EXT('instance').full) {
+      let recreatedUri;
+      if (
+        prop.object.value.includes(
+          'http://data.lblod.info/variables/--ref-uuid4-',
+        )
+      ) {
+        recreatedUri = `http://data.lblod.info/variables/--ref-uuid4-${uuidv4()}`;
```
abeforgit, line 30: would it be better to use the library here? to make sure the behavior is consistent

abeforgit: would also avoid the need for the includes check, so if we ever change the base it would keep working

addon/plugins/variable-plugin/utils/recreate-variable-uris.ts
```diff
@@ -0,0 +1,43 @@
...
+import { sayDataFactory } from '@lblod/ember-rdfa-editor/core/say-data-factory';
+import { OutgoingTriple } from '@lblod/ember-rdfa-editor/core/rdfa-processor';
+import { v4 as uuidv4 } from 'uuid';
+import { Attrs } from '@lblod/ember-rdfa-editor';
+
+export function recreateVariableUris(nodeAttrs: Attrs) {
+  const attrs = { ...nodeAttrs };
+  if (
+    attrs.subject &&
+    (attrs.subject as string).includes(
```
abeforgit, line 11: Same thing, if we use the lib we don't need the includes check here

abeforgit (changes requested):

addon/plugins/variable-plugin/utils/recreate-variable-uris.ts
lagartoverde: What do you mean? the template-uuid-instantiator? Is not what it does, here we recreate a uuid maintaining it being a template or a normal uri. The uuid instantiator transform from template to normal uuid

addon/plugins/variable-plugin/utils/recreate-variable-uris.ts
lagartoverde: done :)

addon/plugins/variable-plugin/utils/recreate-variable-uris.ts
abeforgit: oh yeah sorry, brainfart

abeforgit (changes requested):
think you forgot to configure this for date variables, otherwise looks good

PR #512, opened 2024-11-25
adapt irgn plugin with new model

nbittich:
GN-4997

Still a work in progress, the filters and table seem working, I think template is the next thing to implement once GN-5108 is ready (cf @aliokan )

If you see something I missed, please let me know

uses https://dev-vlag.roadsigns.lblod.info/ for the backend, as it uses the new model.

cf @andreo141 

I've removed `order` as it doesn't seem to be in the new model (I think it will have to be added at some point)

https://github.com/user-attachments/assets/26478e74-31c2-4a39-bac8-4e49ffcd5b8b

### DIFF
--- a/addon/models/sign.ts
+++ b/addon/models/sign.ts
@@ -17,20 +17,15 @@ export default class Sign {
       unwrap(binding['imageBaseUrl']?.value),
     );
 
-    const uri = unwrap(binding['uri']?.value);
-    const order = unwrap(binding['order']?.value);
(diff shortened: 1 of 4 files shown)


aliokan:
Blocked by https://github.com/lblod/fix-annotation-service/pull/8

nbittich:
closed in favour of https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/526

PR #515, opened 2024-12-04
allow user to select how they want the article titles

lagartoverde:
### Overview
Made the user able to select the old article style without a only article title, and always full length articles

GN-5235

### Setup
None

### How to test/reproduce
Change the parameters in besluit config to:
```js
structure: {
        fullLenghtArticles: true,
        disableOnlyArticleSpecialName: true,
},
```
And observe how the style on the article titles change

### Challenges/uncertainties
None

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/addon/components/structure-plugin/_private/structure.gts
+++ b/addon/components/structure-plugin/_private/structure.gts
@@ -120,7 +120,7 @@ export default class Structure extends Component<Sig> {
   get title() {
     if (
       this.node.attrs.isOnlyArticle &&
-      !this.node.attrs.disableOnlyArticleSpecialName
+      this.node.attrs.onlyArticleSpecialName
     ) {
       const docLang = this.controller.mainEditorState.doc.attrs.lang;
       return this.intl.t('structure-plugin.only-article-title', {
--- a/addon/components/structure-plugin/_private/structure.gts
+++ b/addon/components/structure-plugin/_private/structure.gts
@@ -95,8 +95,8 @@ export default class Structure extends Component<Sig> {
     return this.node.attrs.structureType as StructureType;
   }
 
-  get fullLenghtArticles() {
-    return this.node.attrs.fullLenghtArticles as boolean;
+  get fullLengthArticles() {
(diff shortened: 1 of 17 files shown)


addon/components/structure-plugin/_private/structure.gts
```diff
@@ -95,12 +95,17 @@ export default class Structure extends Component<Sig> {
     return this.node.attrs.structureType as StructureType;
   }
 
+  get fullLenghtArticles() {
+    return this.node.attrs.fullLenghtArticles as boolean;
```
elpoelma, line 99: Very pedantic, but this should be `fullLengthArticles`

addon/plugins/structure-plugin/node.ts
```diff
@@ -51,7 +52,14 @@ export type StructureType =
   | 'subsection'
   | 'paragraph';
 
-export const emberNodeConfig: () => EmberNodeConfig = () => {
+type StructureConfig = {
+  fullLenghtArticles: boolean;
+  onlyArticleSpecialName: boolean;
```
elpoelma, line 57: As the `structureConfig` is an optional argument, I would also make both of these fields individually optional.

lagartoverde: good catch :)

addon/components/structure-plugin/_private/structure.gts
abeforgit: oh good catch, not pedantic at all, the typo is also in the config you pass in so this will trip up users

abeforgit (changes requested):
need to fix the lenght -> length typo

addon/components/structure-plugin/_private/structure.gts
lagartoverde: oh sorry brainfarted a bit there

lagartoverde:
> need to fix the lenght -> length typo

thanks missed elena comment

abeforgit (approved):

PR #517, opened 2024-12-09
Add ability to link locations with the wider document automatically

piemonkey:
### Overview
Also move from dct:spatial to prov:atLocation.

I spent some time refactoring the address variable to re-use a load of the code from the oslo location node and so to benefit from these changes, but I realised that substantially increases the scope of the changes. Instead of finishing those changes, I've not committed them. If we think we should include them, I'll add them, otherwise I'll hold on to them for now.

The links are made in RDFa but only the backlink works in the RDFa tools. This is unfixed as we need to refactor how these nodes work anyway, so should do this then.

Edit: In order to avoid a weird situation of having 2 `prov:atLocation` triples with different objects for each location, I included a more substantial refactor which 'de-variable-ifies' the location code. So, now:
- It no longer inserts a placeholder first, then edits that placeholder with the address. We perhaps need an 'insert placeholder' for building templates, but the old behaviour was weird anyway, as it involved inserting a location, then closing the modal without setting it, which is not ideal.
- It no longer inserts a load of extra RDFa for 'mappings' and variable types.
- Supports importing locations from before this change, but does not remove the 'extra' RDFa until the location is edited.

Things this does not include:
- A fix for the 'incorrect RDFa' issue, which is a quick fix, but needs more work to also support address variables.
- Support for template UUIDs in locations. It's unclear whether we want this as we no longer have the 'mappings', but it would be a simple fix to add if we want it.

Jira ticket: https://binnenland.atlassian.net/browse/GN-5022
GN PR: https://github.com/lblod/frontend-gelinkt-notuleren/pull/794
RB PR: https://github.com/lblod/frontend-reglementaire-bijlage/pull/306

### Setup
N/A

### How to test/reproduce
In the besluit or snippet sample pages, insert locations into arbitrary documents, inside Besluits and inside Articles. The expectation (see jira) is that it will link the location to an article if it can find one or a besluit if it cannot. If it can't find either, it will be unlinked.

### Challenges/uncertainties
N/A

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/few-buses-pay.md
+++ b/.changeset/few-buses-pay.md
@@ -1,0 +1,5 @@
+---
+'@lblod/ember-rdfa-editor-lblod-plugins': minor
+---
+
+Add configuration option to location plugin to automatically link to resources in the wider document
--- a/.changeset/five-suits-pull.md
+++ b/.changeset/five-suits-pull.md
@@ -1,0 +1,5 @@
+---
+'@lblod/ember-rdfa-editor-lblod-plugins': minor
+---
+
+Use prov:atLocation predicate for locations instead of dct:spatial
--- a/.changeset/wicked-berries-sing.md
+++ b/.changeset/wicked-berries-sing.md
@@ -1,0 +1,5 @@
+---
+'@lblod/ember-rdfa-editor-lblod-plugins': minor
+---
+
+Remove vestigual 'mapping' RDFa from oslo location nodes as they are no longer 'variables' so no longer need it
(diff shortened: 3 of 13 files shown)


elpoelma (approved):
Looks good to me!
- When inserting a location in a decision/article, it's correctly linked
- The RDFa seems more correct/less verbose now. And seems to be following the data-model correctly.

I do think we'll need a template-version down the road.

PR #518, opened 2024-12-09
Feat/add person annotations

lagartoverde:
### Overview
Add rdfa annotations on the person variable for URI, firstname and lastname

GN-5240

### Setup
None

### How to test/reproduce
Insert a person from LMB and click on show raw export, notice that the person variable now includes info about the uri, firstname and lastname

### Challenges/uncertainties
Not sure if the rdfa generated is correct or could be improved 

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/wet-zebras-run.md
+++ b/.changeset/wet-zebras-run.md
@@ -1,0 +1,5 @@
+---
+'@lblod/ember-rdfa-editor-lblod-plugins': patch
+---
+
+Add rdfa to the person variable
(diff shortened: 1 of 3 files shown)


elpoelma (changes requested):
Makes sense implementation wise overall.
Two things:
- Can you add backwards-compatible `parseDOM` rules to parse older versions of this node?
- An `ext:content` relationship between the mapping and the person URI is probably needed

abeforgit (changes requested):
no space between first name and last name...
image

PR #519, opened 2024-12-11
Fix RDFa output of oslo location addresses

piemonkey:
### Overview
Also deprecates the address variable. Opening a document with an existing location or address variable (assuming the `oslo_location` nodespec comes before the `address` one in the schema, should be loaded as an `oslo_location` and a new URI generated for the address, instead of using the 'belgian address uri'.

PR made to the location linking PR as it didn't make sense to make it separately.

Based on https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/517
Jira ticket: https://binnenland.atlassian.net/browse/GN-5242

### Setup
N/A

### How to test/reproduce
Location variables should now output RDFa which correctly matches [the OSLO model](https://github.com/Informatievlaanderen/OSLOthema-slimmeRaadpleegOmgeving#een-adres), including the fact that the `adres:verwijstNaar` relationship points to the URI given by the basisregisters API but this is no longer also (mis)used as the resource URI.

It also makes sense to test that older versions of this node get parsed correctly.

### Challenges/uncertainties
It was unclear if we still need the address variable in order to provide placeholders, but since we have [a ticket](https://binnenland.atlassian.net/browse/GN-5339) to replace this functionality, this should probably be handled as part of that. I didn't remove the variable as that would make this a breaking change.

A strange thing about the OSLO model for addresses is that in the example, the `adres:verwijstNaar` triple's object is a string, not a URI. This seems wrong, and disagrees with the class diagram. I've changed it to be a URI as that seems more logical.

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations


elpoelma (commented):
Looks good to me.
- Seems to work in a backwards compatible way.
- Maybe we should use `https://publicatie.gelinkt-notuleren.vlaanderen.be/id/adresvoorstelling/` instead of  `https://publicatie.gelinkt-notuleren.vlaanderen.be/id/adres/` ?

lagartoverde (approved):

PR #523, opened 2024-12-27
Redesign of fragment nodes

lagartoverde:
### Overview
Redesigned fragment nodes according to the new design 

GN-5120

### Setup
None

### How to test/reproduce
Insert a snippet, and check that the design of it is the same as in the figma attached to the issue 

### Challenges/uncertainties
Nobody was here so I must have missed something on the design, but didn't see anything more. Also might need to check with saska because I had to recreate the badge which I don't really like

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/fluffy-carrots-drum.md
+++ b/.changeset/fluffy-carrots-drum.md
@@ -1,0 +1,5 @@
+---
+'@lblod/ember-rdfa-editor-lblod-plugins': minor
+---
+
+Redesign of fragment nodes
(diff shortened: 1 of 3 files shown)


piemonkey (commented):
Looks good but there's a few things I think it makes sense to tweak:

- The buttons in the design are at the top rather than at the bottom (though there's a `?` in the comments from Saska)
- The buttons are not centralised in the space, they are closer to the content than the border
- There is a big gap at the start of the snippet

elpoelma (approved):
Looks good to me design-wise, seems to follow the design of Saska well.
- Everything still works as expected
- I like that the buttons are positioned at the top instead of the bottom.

PR #525, opened 2025-01-08
Feat/convert snippet plugin to uris

lagartoverde:
### Overview
The snippet plugin was somehow making weird things converting uris to ids to call the backend, and converting them back to uris, this messed up with uris with a different base, or when the ids didn't match the uris, so I converted everything to just use the uris

GN-5087

### Setup
None

### How to test/reproduce
Play with the snippets and snippets placeholder, verify that everything still works as before

### Challenges/uncertainties
None

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/addon/plugins/snippet-plugin/nodes/snippet.ts
+++ b/addon/plugins/snippet-plugin/nodes/snippet.ts
@@ -35,7 +35,7 @@ import {
   SnippetListProperties,
   type SnippetPluginConfig,
 } from '@lblod/ember-rdfa-editor-lblod-plugins/plugins/snippet-plugin';
-import { tripleForSnippetListId } from '../utils/rdfa-predicate';
+import { tripleForSnippetListUri } from '../utils/rdfa-predicate';
 
 function outgoingFromBacklink(
   backlink: IncomingTriple,
@@ -110,7 +110,7 @@ export function createSnippet({
   });
   const properties = [
     ...defaultProperties,
-    ...listUris.map(tripleForSnippetListId),
+    ...listUris.map(tripleForSnippetListUri),
   ];
   const node = schema.node(
     'snippet',
@@ -214,7 +214,7 @@ const emberNodeConfig = (options: SnippetPluginConfig): EmberNodeConfig => ({
             ? rdfaAttrs.properties
             : [
                 ...rdfaAttrs.properties,
-                ...legacySnippetListUris.map(tripleForSnippetListId),
+                ...legacySnippetListUris.map(tripleForSnippetListUri),
               ];
           return {
             ...rdfaAttrs,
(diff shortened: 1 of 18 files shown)


addon/plugins/snippet-plugin/utils/rdfa-predicate.ts
```diff
@@ -6,33 +6,24 @@ import { getOutgoingTripleList } from '@lblod/ember-rdfa-editor-lblod-plugins/ut
 
 export const SNIPPET_LIST_RDFA_PREDICATE = SAY('allowedSnippetList');
 
-const snippetListBase = 'http://lblod.data.gift/snippet-lists/';
-
-export const getSnippetUriFromId = (id: string) => `${snippetListBase}${id}`;
-
-export const getSnippetIdFromUri = (uri: string) =>
```
piemonkey, line 13: This function is still used in the `updateSnippetPlaceholder` command.

addon/plugins/snippet-plugin/nodes/snippet.ts
```diff
@@ -207,14 +207,14 @@ const emberNodeConfig = (options: SnippetPluginConfig): EmberNodeConfig => ({
...
-          const legacySnippetListIds = node
+          const legacySnippetListUris = node
             .getAttribute('data-assigned-snippet-ids')
             ?.split(',');
-          const properties = !legacySnippetListIds
+          const properties = !legacySnippetListUris
             ? rdfaAttrs.properties
             : [
                 ...rdfaAttrs.properties,
-                ...legacySnippetListIds.map(tripleForSnippetListId),
```
piemonkey, line 217: `tripleForSnippetListId` is not the same as `tripleForSnippetListUri`, but you're passing it the same argument (even if that arg has changed name). This means that it will try to construct named nodes pointing to the id, not the URI. We need to still support the conversion of ids to uris for backwards compatibility.

piemonkey (changes requested):
Mostly seems to work, but the 'manage snippet lists' button is broken due to the command not being updated. It looks like backwards compatibility could be a problem too.

addon/plugins/snippet-plugin/utils/rdfa-predicate.ts
lagartoverde: Solved that, good catch :)

addon/plugins/snippet-plugin/nodes/snippet.ts
lagartoverde: You are right, I couldn't manage to test this, as I cannot form a snippet with ids on the properties, but it should work now, my bad

piemonkey: The change was introduced [here](https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/495), so creating a doc with snippets on <v25.0.0 and then trying to load it on this branch is the way to test. It seems to work.

piemonkey (approved):
Looks good to me.

PR #530, opened 2025-01-17
Move to builtin ember types

elpoelma:
### Overview
This PR moves away from the old, deprecated DefinitelyTyped packages.
Some slight code changes were required.

### How to test/reproduce
- Ensure `pnpm lint:types` is successful

### Checks PR readiness
- [x] Check database state correct when deleting/updating (especially regarding relationships)
- [x] changelog
- [x] npm lint
- [x] no new deprecations


addon/components/location-plugin/edit.gts
```diff
@@ -35,6 +34,7 @@ import {
 import { type Point } from '@lblod/ember-rdfa-editor-lblod-plugins/plugins/location-plugin/utils/geo-helpers';
 import { type NodeContentsUtils } from '@lblod/ember-rdfa-editor-lblod-plugins/plugins/location-plugin/node-contents';
 import { type LocationType } from './map';
+import { SafeString } from '@ember/template';
```
piemonkey, line 37: I can't remember if there was an advantage to using `import type`. We should probably align on one way to do it.

piemonkey (approved):

addon/components/location-plugin/edit.gts
elpoelma: I think the main advantage is clarity and the fact you can prevent import cycles.

elpoelma: There is an eslint rule for it: https://typescript-eslint.io/rules/consistent-type-imports/

PR #532, opened 2025-01-28
fix: prevent `write-after-read` error in codelist and location edit components

elpoelma:
### Overview
This PR includes a workaround/fix to prevent the  'write-after-read' issues when opening the location/codelist edit components.
The 'write-after-read' error starts to popup after an update to ember 5.6+.
The error does not occur on deployed environments, as there 'write-after-read' errors are not thrown...

**Update**
I applied a solution where `trackedFunction` is not used, and the `selected` property is reset in the `trackedFunction`. This also seems to work well.

### Setup
Ensure you're testing this PR (and comparing to master) with a ember-source version of 5.6 or higher.

### How to test/reproduce
**Using master**
- Start the dummy app
- Insert a codelist/classic location variable in any way
- Click on it
- The card with the dropdown appears, and a 'write-after-read' error is thrown
- The dropdown is unusable, as is the rest of the application

**With this PR**
- Start the dummy app
- Insert a codelist/classic location variable in any way
- Click on it
- No 'write-after-read' error is thrown

### Challenges/uncertainties
- This is not an ideal solution, but merely a type of workaround. It introduces side-effects in the `trackedFunction`
- The error seems to be thrown only when using `tracked-toolbox` in combination with e.g. a `trackedFunction`
- (As the data typically loads quickly, the loading flashing is not ideal for the UX of users. We might want to introduce some artificial delay...)

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/addon/components/variable-plugin/codelist/edit.ts
+++ b/addon/components/variable-plugin/codelist/edit.ts
@@ -84,6 +84,9 @@ export default class CodelistEditComponent extends Component<Args> {
     if (this.source && this.codelistUri) {
       result = await fetchCodeListOptions(this.source, this.codelistUri);
     }
+    // This a workaround/hack to be able to reset the `selected` option after the `codelistOptions` change.
+    // Normally we'd do this with a `trackedReset`, but this gave us `write-after-read` dev-errors at the time of writing this.
+    // TODO: convert this back to a `trackedReset` (or an alternative) once possible.
     this.selectedCodelistOption = undefined;
     return result;
   });
--- a/addon/components/variable-plugin/codelist/edit.ts
+++ b/addon/components/variable-plugin/codelist/edit.ts
@@ -3,12 +3,12 @@ import { action } from '@ember/object';
 import { SayController } from '@lblod/ember-rdfa-editor';
 import {
   CodeListOption,
+  CodeListOptions,
   fetchCodeListOptions,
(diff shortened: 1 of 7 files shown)


addon/components/variable-plugin/codelist/edit.ts
```diff
@@ -81,11 +80,12 @@ export default class CodelistEditComponent extends Component<Args> {
...
 
   codelistOptions = trackedFunction(this, async () => {
+    let result: CodeListOptions | undefined;
     if (this.source && this.codelistUri) {
-      return fetchCodeListOptions(this.source, this.codelistUri);
-    } else {
-      return;
+      result = await fetchCodeListOptions(this.source, this.codelistUri);
     }
+    this.selectedCodelistOption = undefined;
```
abeforgit, line 87: I'd like a comment here to explain this (= resources with side-effects) is really not a pattern we should internalize, and it's basically a hack that nullvox gave us

abeforgit (changes requested):

addon/components/variable-plugin/codelist/edit.ts
abeforgit: otherwise good to go

elpoelma: Good point.

PR #535, opened 2025-02-05
Move inserting article structures to use structure nodes

piemonkey:
### Overview
Move article structures to use the 'structure' nodespec instead of using different nodespecs for each structure. There are pieces missing from this in order to break the work into pieces...
- [x] [Moving of structures](https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/537)
- [x] [Setting of 'start from' for structure numbers](https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/537)
- ~~[ ] Improvements to deleting structures to make it possible without having to select a block of text around the structure~~ Given the article structure card works to delete structures and so the options to delete them are the same as for the old version, I think we can skip this for now.
- [x] Improvements to the design of structures with titles
- [x] [Parsing of old article-structure-plugin nodes as structure nodes](https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/538)
- [x] [Ensuring that RDFa output is correct](https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/538)
- [x] [Removing the now unused code](https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/539)

Requires https://github.com/lblod/ember-rdfa-editor/pull/1249
Jira ticket: https://binnenland.atlassian.net/browse/GN-4866

### Setup
N/A

### How to test/reproduce
In the regulatory statements sample page it should be possible to insert article structures in a way that makes sense...

### Challenges/uncertainties
This was a mess, I don't know why it took so long, but the use of structure nodespecs for decisions felt like it was built on a load of specific hacks, which didn't help... This now feels similarly built on hacks, but at least it includes article structures.

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations


addon/plugins/structure-plugin/insert-structure.ts
```diff
@@ -0,0 +1,136 @@
...
+  structureType: StructureType,
+  uriGenerator: StructurePluginOptions['uriGenerator'],
+): Transaction | false {
+  const structConfig = STRUCTURE_HIERARCHY.find(
+    ({ structureType: type }) => type === structureType,
+  );
+  const hierarchyRank =
+    structConfig && findRankInHierarchy(structConfig.rdfType.full);
+  if (!structConfig || isNone(hierarchyRank)) {
+    console.warn('not a structure', structureType);
```
lagartoverde, line 42: Maybe improve this error message a little bit :)

lagartoverde (commented):
Code looks good, some minor comments, will test it now

addon/components/variable-plugin/date/edit.ts
```diff
@@ -277,7 +277,7 @@ export default class DateEditComponent extends Component<Args> {
         return tr.setNodeAttribute(pos, 'custom', true);
       });
     } else {
-      const format = this.formats.find((format) => format.key === formatKey);
+      const format = this.formats.find((frmt) => frmt.key === formatKey);
```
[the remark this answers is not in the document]
piemonkey: This is due to the 'no shadow' rule I enabled. I agree `format` is clearer, but it also leaves the space for confusion between the `format` in the arrow function and the `format` we're assigning. I thought it was clear enough for this one-liner without having to either come up with a longer clear name or disabling the rule.

addon/plugins/structure-plugin/insert-structure.ts
piemonkey: Good point. Fixed.

abeforgit:
Had a look and a try, seems to work very nicely overall
already more flexible than it ever was, and it keeps the hierarchy

I think the code looks clean too, what you call hacks I just call editor code tbh :grimacing: I think anything non-trivial inevitably ends up extremely branch-heavy 

one thing I can maybe take over for you is resolving the selection behavior in the title, that seems fairly isolated
and perhaps @lagartoverde can help by re-implementing the deletion logic?

abeforgit:
@piemonkey think I have a pretty clean solution for the title behavior, have a look: https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/536

elpoelma:
Looks really good UX/UI wise, I did find a bit of a weird bug when wrapping structures in a title-structure:

https://github.com/user-attachments/assets/ebfe22f1-2c9b-4cf2-8253-b0957e81ca78

elpoelma:
Still found another issue:

Uploading Screencast From 2025-02-20 11-12-21.mp4…

elpoelma (approved):
One the editor is released, this PR is good to merge for me! Works really well! :)

elpoelma:
@piemonkey editor is released and updated, so I think we can merge this PR.

PR #536, opened 2025-02-10
fix(structures): add css solution to fix title behavior

abeforgit:
### Overview

Makes clicking in titles work better by leveraging flex rendering to make the title box as wide as the whole header

also removes the margin from the content, which came from a rule somewhere in the editor stylings. I used important here to get around it, which is not ideal but it works.

This prevents the state where clicking in just the wrong spot node-selects the structure (characterized by having no blinking cursor), after which typing would replace the whole structure with the typed text.

Another way to prevent this would have been making the structure no longer node-selectable, but that then interferes with how we likely want the backspace into the structure nodes to behave.

built on #535 

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/app/styles/structure-plugin.scss
+++ b/app/styles/structure-plugin.scss
@@ -31,7 +31,6 @@
     .say-structure__title {
       flex: 1 1 100%;
 
-
       border: none;
       flex-grow: 1;
       display: inline-block;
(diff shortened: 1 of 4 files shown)


app/styles/structure-plugin.scss
```diff
@@ -6,39 +6,76 @@
...
+
+  h6.say-structure__header-element {
+    font-size: var(--au-h6);
   }
 }
+
 .say-structure__content {
   padding: 0.7rem;
   padding-right: 3rem;
+  margin-top: 0 !important;
```
piemonkey, line 77: We can get around this by using `.say-structure > .say-structure__content` to increase the specificity.

piemonkey (commented):
Looks good, would be good to get rid of the `!important` though.

piemonkey:
Actually, since I have a PR I want to open from the same branch, maybe I'll make the `!important` tweak, merge to the branch and then rebase mine, to avoid the psychic damage of having diverging branches.

PR #537, opened 2025-02-10
Add support for moving and numbering structures to structure plugin

piemonkey:
### Overview
The structure plugin code for this was generic enough that this was easy, just needed some tweaking and translations. ~~Still need to include the 'start number' functionality to replace the article structure card.~~

For reviewing, I recommend the 'hide whitespace changes' option.

##### connected issues and PRs:
Part of https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/535
Jira ticket: [binnenland.atlassian.net/browse/GN-4866](https://binnenland.atlassian.net/browse/GN-4866)

### Setup
N/A

### How to test/reproduce
The card to move and remove structures in the RS part of the dummy app should work as expected.
Setting start numbers should work as expected.

### Challenges/uncertainties
N/A

### Checks PR readiness
- [ ] UI: works on smaller screen sizes
- [ ] UI: feedback for any loading/error states
- [ ] Check if dummy app is correctly updated
- [ ] Check cancel/go-back flows
- [x] changelog
- [x] npm lint
- [x] no new deprecations

PR #538, opened 2025-02-17
Article structure parse rule and RDFa output

piemonkey:
### Overview
Parses article-structure-plugin nodes as structures. Adds RDFa links to structure nodes that link the title and body to the structure. These match those used for article-structure nodes but also apply to decision articles too now.

Part of https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/535
Along with: https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/537
Jira ticket: [binnenland.atlassian.net/browse/GN-4866](https://binnenland.atlassian.net/browse/GN-4866)

### Setup
N/A

### How to test/reproduce
Use an older version to generate some exported HTML, e.g. dev GN, then paste it into the plugins test app and see what happens to the article structures.
Check in a tool, (such as [rdfa play](https://rdfa.info/play/) or [is semantic](https://issemantic.net/rdf-visualizer) that the RDFa for both matches, with the exception of the `say:heading` triple which is no longer used.

### Challenges/uncertainties
It turns out the old nodes were a bit broken in places...

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations


piemonkey:
I'll merge this. Any last comments will need to be in https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/535

PR #540, opened 2025-02-19
Fix structure rendering and history bugs

abeforgit:
### Overview

Fixes 4 bugs:
- case where dom state and editor state could diverge (see https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/535#issuecomment-2656482078)
- fixes spurious undo entries when inserting a structure
- fixes spurious undo entries when editing a title
- adds ctrl-z/y keymap to the nested header editor
### Checks PR readiness

- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/addon/components/structure-plugin/_private/structure.gts
+++ b/addon/components/structure-plugin/_private/structure.gts
@@ -26,6 +26,7 @@ import IntlService from 'ember-intl/services/intl';
 import { getNameForStructureType } from '@lblod/ember-rdfa-editor-lblod-plugins/plugins/structure-plugin/node';
 import { StructureType } from '@lblod/ember-rdfa-editor-lblod-plugins/plugins/structure-plugin/structure-types';
 import { romanize } from '@lblod/ember-rdfa-editor-lblod-plugins/plugins/article-structure-plugin/utils/romanize';
+import { Transaction } from '@lblod/ember-rdfa-editor';
 
 interface Sig {
   Args: EmberNodeArgs;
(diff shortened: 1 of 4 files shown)


addon/components/structure-plugin/_private/structure.gts
```diff
@@ -148,8 +154,21 @@ export default class Structure extends Component<Sig> {
...
+    //
+    // but since we trigger an attribute update, the history plugin can no
+    // longer recognize these edits as adjacent, so each character triggers
+    // a history event. This makes undoing title edits tedious, so we mimic the
+    // grouping here using the transaction timestamp
+    if (tr.time - this.historyTimeStamp > 500) {
+      addToHistory = true;
+      this.historyTimeStamp = tr.time;
+    }
+    this.args.updateAttribute('title', content, !addToHistory);
```
elpoelma, line 171: It doesn't really matter here, as these are attribute updates, but I'd think using the `composition` meta here is more correct than the `addToHistory` meta. `addToHistory: false` transactions can never be undoed, if this transaction here would (at one instance) contain another document change, this change would never be able to be undoed...

elpoelma (changes requested):
Looks really good overall! Just have one comment on the adjacent-edit stategy.

piemonkey (approved):
Looks good to me, thanks for looking into this mess.

addon/components/structure-plugin/_private/structure.gts
abeforgit: Idk about "more correct", "composition" is a private and undocumented api after all
and given we check the timestamps, the most history you can lose is 500ms worth of edits

elpoelma: Agreed that it's not ideal as it indeed seems to be a private API. 
What I mainly mean with 'incorrect' is the following (in case of real document changes):

https://github.com/user-attachments/assets/660d528f-d19f-44f8-928f-e93ace6084c8

Plugins which react with document changes are in this case also not undoed, this might sometimes be a bit hard to debug.
So in this example it's not really a problem. I just think we should be a bit careful around using the `addToHistory` meta.

PR #543, opened 2025-02-20
fix: handle case where structure title is `null`

elpoelma:
### Overview
This PR passes an empty string to the `embedded` title editor in case the structure title is `null`.

### How to test/reproduce
- Open the RB page
- Insert some structures
- Ensure that no errors show up related to the structure title being `null`


piemonkey (approved):

lagartoverde (approved):
works :)

PR #546, opened 2025-02-24
GN-5468: add support for configuring location types

elpoelma:
### Overview
This PR adds support for configuring the location types shown by the location plugin.
The location types may be configured through the `@locationTypes` argument of the `location-plugin/insert` component.

The `@locationTypes` argument expects an array of `LocationType` values.
Supported `LocationType` values are:
- 'address'
- 'place'
- 'area'
By default the value of `@locationTypes` is `['address', 'place', 'area']`

[GN-5468](https://binnenland.atlassian.net/browse/GN-5468)

### How to test/reproduce
- Start the dummy app
- Open the 'insert location' modal
- Notice that the order of the tabs has changed (in correspondence with the config inside the dummy app)
- If you adjust the `@locationTypes` argument, the modal should update accordingly
- Passing and unsupported location type causes an error. 

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/addon/components/location-plugin/edit.gts
+++ b/addon/components/location-plugin/edit.gts
@@ -109,16 +109,6 @@ export default class LocationPluginEditComponent extends Component<Signature> {
   newTruncatedValue?: boolean;
 
   get locationTypes() {
-    const unsupportedLocationType = this.args.locationTypes?.find(
-      (locationType) => !SUPPORTED_LOCATION_TYPES.includes(locationType),
-    );
-    if (unsupportedLocationType) {
-      throw new Error(
-        `${unsupportedLocationType} is not supported. Supported location types are ${SUPPORTED_LOCATION_TYPES.join(
-          ', ',
-        )}`,
-      );
-    }
     return this.args.locationTypes ?? SUPPORTED_LOCATION_TYPES;
   }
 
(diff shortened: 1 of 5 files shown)


addon/components/location-plugin/edit.gts
```diff
@@ -107,6 +108,31 @@ export default class LocationPluginEditComponent extends Component<Signature> {
   })
   newTruncatedValue?: boolean;
 
+  get locationTypes() {
+    const unsupportedLocationType = this.args.locationTypes?.find(
+      (locationType) => !SUPPORTED_LOCATION_TYPES.includes(locationType),
+    );
+    if (unsupportedLocationType) {
+      throw new Error(
```
piemonkey, line 116: I'm not sure about throwing an error in the edit component, as it only shows the error when the component is shown, and completely crashes the editor when it is. I feel like it would be better to check for this in the insert component, as that way a misconfiguration is immediately spotted. Alternatively, we could log/show an error and continue gracefully, but I feel like this could also hide misconfigurations in practice.

piemonkey (commented):
Looks good, my only doubt is about when to show the error

addon/components/location-plugin/edit.gts
elpoelma: I'll probably move it to the constructor then, not ideal, but putting it in the getter does not solve the issue, as it is only run when it's actually being used in the template...

piemonkey: I'd say constructor is fine, since it's to catch misconfiguration. It's not like we'd expect these to change dynamically.

elpoelma: True true...

abeforgit (approved):

PR #547, opened 2025-02-26
Rework ToC to be compatible with new structure nodes

abeforgit:
### Overview

BREAKING: this bumps the peerdeps of ember-appuniversum and ember-concurrency to the ranges the editor needs. 

- the config is no longer required. Instead, the ToC looks at the `tocEntry` property of a nodeSpec. This makes the ToC logic way simpler and decouples it from the implementation of hierarchies and other complexities
- Passing in the legacy config, an array, will active the legacy mode. So the behavior will not change until you change the config.
- passing in an object (not an array) with a `scrollContainer` callback is now also supported, but it is entirely optional as the plugin will try to find its scrollcontainer on its own.

https://binnenland.atlassian.net/browse/GN-5387

### How to test/reproduce

go to the regulatory statement sample page, add some structures, add the ToC.
Observe it automatically updates when you make changes to the structures. (including editing their titles)

### Challenges/uncertainties

Known issue: you have to click twice on the TOC entry to jump to the position in the document. I don't think that's a regression from this PR, judging by this ticket: https://binnenland.atlassian.net/browse/GN-4419 
It's a tricky one to solve, so I'd rather unblock the plugin as a whole first and pick up that minor ux inconvenience later.

### Checks PR readiness

- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/addon/plugins/structure-plugin/node.ts
+++ b/addon/plugins/structure-plugin/node.ts
@@ -39,11 +39,9 @@ import {
 } from './structure-types';
 import { parseBooleanDatasetAttribute } from '@lblod/ember-rdfa-editor-lblod-plugins/utils/dom-utils';
 import {
-  isSome,
   Option,
   optionMapOr,
 } from '@lblod/ember-rdfa-editor-lblod-plugins/utils/option';
-import { unwrapOr } from '@lblod/ember-rdfa-editor/utils/_private/option';
 
 const rdfaAware = true;
 
@@ -162,7 +160,6 @@ export const emberNodeConfig: (
       },
     },
     serialize(node: PNode, state: EditorState) {
-      const parser = new DOMParser();
       const tag = node.attrs.headerTag;
       const structureType = node.attrs.structureType as StructureType;
       const number = node.attrs.number as number;
--- a/addon/plugins/structure-plugin/node.ts
+++ b/addon/plugins/structure-plugin/node.ts
@@ -82,6 +82,9 @@ export const emberNodeConfig: (
     isolating: true,
     atom: false,
     editable: rdfaAware,
+    tocEntry: (node: PNode) => {
+      return `${node.attrs.number}. ${node.attrs.title}`;
+    },
 
     attrs: {
       ...rdfaAttrSpec({ rdfaAware }),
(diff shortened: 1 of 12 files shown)


addon/plugins/table-of-contents-plugin/utils/index.ts
```diff
@@ -43,17 +43,64 @@ export function createTableOfContents(entries: OutlineEntry[]) {
   return tableOfContents;
 }
 
-type Config = { nodeHierarchy: string[] }[];
+type LegacyConfig = { nodeHierarchy: string[] }[];
+// the scrollcontainer config isn't used here, so we can use the unknown type
+// cause we don't need any config
+type NewConfig = unknown;
```
piemonkey, line 49: I get that we don't need anything from the config here, but using unknown means we're losing any type checking on the config when calling `extractOutline()`. Why not just use `TableOfContentsConfig`?

piemonkey (commented):
Looks good except for the `null` for paragraphs.

I also don't know if it was an existing problem, but trying to jump to a structure seems to jump to the most nested structure in that section. For example if I have chapter 3 > section 1 > article 1, and I click on 'chapter 3', the editor jumps to have the cursor on 'article 1' and with the cursor just at the top of the screen, so I don't actually see 'chapter 3'. The scrolling in general seems a bit broken as I was able to get it to either not move the cursor or get 'out of range' errors by clicking around.

addon/plugins/structure-plugin/node.ts
```diff
@@ -82,6 +86,40 @@ export const emberNodeConfig: (
...
     editable: rdfaAware,
+    tocEntry: (node: PNode, state: EditorState) => {
+      const intlService = getIntlService(state);
+
+      const structureType = node.attrs.structureType as StructureType;
+      const number = node.attrs.number as number;
+
+      const headerFormat = node.attrs.headerFormat as string;
+      const structureName =
+        headerFormat === 'name'
```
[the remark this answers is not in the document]
abeforgit: good catch, on me for not testing all the structures :see_no_evil:

addon/plugins/table-of-contents-plugin/utils/index.ts
abeforgit: yeah fair, I'll change it

PR #555, opened 2025-03-24
feat: add backward-compatible parserule to `roadsign_regulation`  nodespec

elpoelma:
### Overview
This PR adds a backward-compatible parserule to the `roadsign_regulation` nodespec which converts classic `roadsign_regulation` nodes to `block_rdfa` nodes in the best of its capabilities.

None

### How to test/reproduce
- Start the test-app and open the decision sample page
- Find an (older) 'reglement' decision on GN, which contains a classic `roadsign_regulation` node
- Copy the decision and paste it in the test-app
- Ensure the `roadsign_regulation` node has been converted to a `block_rdfa` node and contains the necessary RDFa properties and relationships.

### Challenges/uncertainties
- Adding the parserule to the `roadsign_regulation` spec itself is not ideal. I'd prefer using some kind of 'loose' parserule. It feels a bit silly to just add a nodespec just for its parserule...
- Unsure if we should mark this as breaking
- Hopefully the converted node contains all necessary RDFa

### Checks PR readiness
- [x] changelog
- [x] npm lint

### DIFF
--- a/addon/plugins/roadsign-regulation-plugin/nodes.ts
+++ b/addon/plugins/roadsign-regulation-plugin/nodes.ts
@@ -6,16 +6,17 @@ import {
   PROV,
 } from '@lblod/ember-rdfa-editor-lblod-plugins/utils/constants';
 import { hasRDFaAttribute } from '@lblod/ember-rdfa-editor-lblod-plugins/utils/namespace';
+import { sayDataFactory } from '@lblod/ember-rdfa-editor/core/say-data-factory';
 
 const CONTENT_SELECTOR = `div[property~='${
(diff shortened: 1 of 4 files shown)


addon/plugins/roadsign-regulation-plugin/nodes.ts
```diff
@@ -61,7 +62,12 @@ export const roadsign_regulation: NodeSpec = {
   parseDOM: [
     {
       tag: 'div',
+      node: 'block_rdfa',
```
abeforgit, line 65: whoa we can do this?

PR #558, opened 2025-03-27
Use correct rdf:value instead of ext:content when serializing

abeforgit:

[GN-5539](https://binnenland.atlassian.net/browse/GN-5539)
### Challenges/uncertainties

Not 100% sure all old variables will get parsed into a state where they use rdf:value
I checked the parserules and it seems so, though the "modern" one just takes the rdfa attributes verbatim so there's a slight chance it could get it wrong, but I don't think so

### Checks PR readiness

- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/itchy-coats-nail.md
+++ b/.changeset/itchy-coats-nail.md
@@ -1,0 +1,5 @@
+---
+'@lblod/ember-rdfa-editor-lblod-plugins': patch
+---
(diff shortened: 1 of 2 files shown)


elpoelma (approved):
Looks good! This should be backward compatible.

PR #564, opened 2025-04-17
Feat/make lpdc endpoint more configurable

lagartoverde:
### Overview
Made the ldpc endpoint more configurable so we can fetch the concepts in any app if needed to

CIT-27

### Setup
None

### How to test/reproduce
Check that the lpdc plugin still works as before

### Challenges/uncertainties
Tried to do this non-breaking and support old configs, but it was more of a headache than it was worth, we will need to communicate with the people using the plugin if any

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/gorgeous-rocks-cheer.md
+++ b/.changeset/gorgeous-rocks-cheer.md
@@ -1,0 +1,26 @@
+---
+'@lblod/ember-rdfa-editor-lblod-plugins': major
+---
+
+BREAKING: changed the endpoint config on lpdc plugin so you can specify if you want instances or concepts.
+The config of the lpdc plugin should go from: 
+```
+lpdc: {
+  endpoint: '/lpdc-service',
+},
+```
+ to
+
+```
+lpdc: {
+  endpoint: '/lpdc-service/doc/instantie',
+},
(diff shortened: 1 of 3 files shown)


piemonkey (commented):
Couldn't we instead of changing the meaning of `endpoint`, add an optional field to the config for `path`? This could then default to `/doc/instantie`. Alternatively, we could use a `type` field that we can set to either `concepts` or `instances`, but I think I prefer the first one.

lagartoverde:
> Couldn't we instead of changing the meaning of `endpoint`, add an optional field to the config for `path`? This could then default to `/doc/instantie`. Alternatively, we could use a `type` field that we can set to either `concepts` or `instances`, but I think I prefer the first one.

I thought about this, but I don't think this accomplish anything better than just changing the meaning of endpoint. Having 2 different configs for the same endpoint is not something I would want tbh and makes everything more complex if the final user has a different structure in his proxy than ours

piemonkey (approved):
After discussion, it makes sense to keep this clean as we don't need to care too much about the breaking release.

PR #568, opened 2025-04-23
HOTFIX (28.x): Instantiate template uris in standard template plugin

abeforgit:
### Overview

A single forgotten function call which caused a huge mess in terms of bad publications...

- replaces the templates used in the dummy app with the current ones in the GN database (this was probably one of the reasons we didn't notice the problem)
- runs the template-uuid-instantiator when inserting a template with the standard-template-plugin

https://binnenland.atlassian.net/browse/GN-5572

### How to test/reproduce

commit ab6935c2dbc94ffd8b5c5a319e082bd005100eab reproduces the bug in the dummy app by using the current GN templates instead of the old ones we still had
(insert a decision template using one of the sidebar buttons, notice the decision still has a `--ref-uuid4` uri)

commit d8bc81ff631e5a62f77e217c2ffcce0a656cd863 fixes the bug by running the template-uuid-instantiator.

Alternatively, hook up to gn and test creating an agendapoint from the meeting screen (which doesn't use the new template selector)

### Checks PR readiness
- [ ] changelog
- [ ] npm lint
- [ ] no new deprecations

### DIFF
--- a/.prettierrc.js
+++ b/.prettierrc.js
@@ -3,4 +3,5 @@
 module.exports = {
   plugins: ['prettier-plugin-ember-template-tag'],
   singleQuote: true,
+  htmlWhitespaceSensitivity: 'strict',
 };
--- a/addon/components/standard-template-plugin/template-provider.ts
+++ b/addon/components/standard-template-plugin/template-provider.ts
@@ -13,6 +13,7 @@ import {
   hasOutgoingNamedNodeTriple,
   Resource,
 } from '@lblod/ember-rdfa-editor-lblod-plugins/utils/namespace';
+import templateUuidInstantiator from '@lblod/template-uuid-instantiator';
 
 type Args = {
   controller: SayController;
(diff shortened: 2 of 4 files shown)


piemonkey (approved):
It looks like the `InsertArticleComponent` on the test app is incorrectly configured to insert `--ref-uuid4` uris, but unless that's also the case on GN I can't think of any other places we should be instantiating these URIs.

In terms of future fixes, could we also run the instantiator in the prepublisher to make sure anything in the output document is correctly set up?

abeforgit:
> It looks like the `InsertArticleComponent` on the test app is incorrectly configured to insert `--ref-uuid4` uris, but unless that's also the case on GN I can't think of any other places we should be instantiating these URIs.

good thinking, I'll doublecheck

> In terms of future fixes, could we also run the instantiator in the prepublisher to make sure anything in the output document is correctly set up?

Yeah I was thinking that as well, but there's only a limited amound of documents we can do this. For the notulen for example: if multiple decisions have the same template uri (which they would), they would all get replaced by the same real uri, which is just as bad basically

abeforgit:
> It looks like the `InsertArticleComponent` on the test app is incorrectly configured to insert `--ref-uuid4` uris, but unless that's also the case on GN I can't think of any other places we should be instantiating these URIs.

In GN we configure it correctly indeed. However I'm thinking now we shouldn't make the `--ref` uri the default generation of the articles, I think a real uri is safer. 
That being said, because article uris are always generated, it's also less of an issue, a `--ref` uri, while a bit ugly, is on its own a valid uri, so as long as its uuid is still unique, it's ok (although of course to be avoided)

PR #577, opened 2025-05-28
Support drag-drop of structures and snippets

piemonkey:
### Overview
Styling and behavour changes to support selecting and drag-drop of structures and snippets. Similar to rdfa blocks, they can be selected by clicking on the title. This is made more complex by the nested prosemirror in structure titles.

The on-changed plugin is used to make sure structure numbers and links are kept up to date whenever they are moved.

There is still one weird behaviour with selecting structures, which is in many cases (it depends what's currently selected), when selecting a structure, the cursor jumps inside the structure, before selecting the node itself. This can be fixed with a `preventDefault()` in the structure's `stopEvent()`, similar to the one for snippets, but this has the unfortunate side-effect that if a structure title anywhere in the document is selected, when any structure name is clicked, the selection remains where it is instead of moving. Due to the complexity of detecting this situation it seems better to leave the weird visual bug in place.

Part of [binnenland.atlassian.net/browse/GN-5579](https://binnenland.atlassian.net/browse/GN-5579)
Editor PR: https://github.com/lblod/ember-rdfa-editor/pull/1321

### Setup
Already has the relevant editor version included.

### How to test/reproduce
Move structures, snippets and RDFa blocks around by dragging and dropping. Care should be taken to test different combinations of nesting and where the selection is before the process starts.
Verify that structures are kept up to date in terms of numbering and RDFa links.
All existing interactive functionalities of snippets, structures and RDFa blocks should remain intact.

### Challenges/uncertainties
This was way more fiddly than expected. There seems to be a lot competing for any events...

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/early-badgers-sell.md
+++ b/.changeset/early-badgers-sell.md
@@ -1,0 +1,5 @@
+---
+'@lblod/ember-rdfa-editor-lblod-plugins': minor
+---
+
+Move structure-plugin to use on-changed plugin to always keep RDFa and numbering up to date on changes
--- a/.changeset/plenty-rockets-tell.md
+++ b/.changeset/plenty-rockets-tell.md
@@ -1,0 +1,5 @@
+---
+'@lblod/ember-rdfa-editor-lblod-plugins': minor
+---
+
+Move to @lblod/ember-rdfa-editor version with support for drag-and-drop ember-nodes and on-changed plugin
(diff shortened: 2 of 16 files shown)


elpoelma (commented):
Seems to work really nicely. Would it be hard to extend the configuration to also allow for drag-and-dropping snippet nodes?

elpoelma (approved):
Works nicely now with snippets!

PR #579, opened 2025-06-05
Change structure-plugin node body to rdf:HTML datatype

piemonkey:
### Overview
The body isn't (necessarily) valid XML, so use the more correct rdf:HTML datatype. This prevents problems with parsers failing if they try to parse invalid XML, such as unclosed self-closing elements, which are valid HTML.

Jira ticket: https://binnenland.atlassian.net/browse/GN-5623

### Setup
Test setup is a little tricky to actually parse the documents, I'll add some PRs with additional docs on how to set up publisher and harvester locally...

### How to test/reproduce
Create a document with structures (a besluit or a RS) and export the HTML. Put the HTML into an RDFa parser, such as rdfa.org/play and verify that the `say:body` has the `rdf:HTML` datatype.

### Challenges/uncertainties
Getting the test setup working...

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/chatty-beers-do.md
+++ b/.changeset/chatty-beers-do.md
@@ -1,0 +1,5 @@
+---
+'@lblod/ember-rdfa-editor-lblod-plugins': minor
+---
+
+Change structure-plugin to output say:body with datatype rdf:HTML instead of rdf:XMLLiteral as it is more correct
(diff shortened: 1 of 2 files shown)


elpoelma (approved):
Tested this with the new experimental rdfa parser. The fix seems to work well!

PR #584, opened 2025-06-19
location-plugin: add placeholder mode

elpoelma:
### Overview
This PR adds a simple placeholder mode to the oslo_location plugin.
It adds a simple sidebar widget which can be used to insert a location placeholder.
Additionally, the serialization of the location node has been adapted to include the `data-say-variable` and `data-say-variable-type` attributes, similar to the approach of other variables.

### Challenges/uncertainties
- Unsure about the uri used for the location placeholder node
- (might be breaking, there is maybe a better way) I had to adapt the signature of the `replaceSelectionWithLocation` function a bit to accept a seperate subject/uri and to make the `toInsert` argument optional.

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations


lagartoverde (approved):

elpoelma:
> > Will remove these, on the change signature of the `replaceSelectionWithLocation`, do you think we should do this differently? As it is only used internally afaict. Maybe we should move a copy to a `_private` folder and mark the original as deprecated?
> 
> Hmm, yes, I think this makes sense, especially as the deprecated one can trivially call the new one. It should probably be an internal API to be honest.
> 
> Maybe we could do with an explicit way off commenting things 'to be removed on next major'...

Yeah that would be nice indeed. I added a _private folder and move the utility functions over there, while still re-exporting them (for now) in the original files.

PR #589, opened 2025-07-07
Add actions to document validation plugin

lagartoverde:
### Overview
 This PR allows the user to add a function to fix a shacl violation, like add a title if there isn't one. For this we have several changes:
 First I added uris to the shacl properties, I made this change in the validation shapes library but also copied it here as it's needed for this to work, when it gets merged there I will remove the shape from here and just import it.
Then I added a file on the plugin with common fixes like the ones we had in the decision plugin
And finally developed a new API to add these actions to the plugin config

GN-5660

### Setup
None
### How to test/reproduce
Check the added actions work, but most important check the API and see if it makes sense

### Challenges/uncertainties
I don't know what to do with the description, because the shacl rule we have in the model says that a besluit can have 0 descriptions and be completely valid, so the fix to add a description doesn't work, as a besluit with 0 description would be valid.
Maybe we can add warning shacl rules? so the besluit is valid with 0 description but it gives a warning, but this can be done in future tickets I think

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/addon/components/document-validation-plugin/card.gts
+++ b/addon/components/document-validation-plugin/card.gts
@@ -41,7 +41,7 @@ export default class DocumentValidationPluginCard extends Component<Sig> {
     if (!propertiesWithErrors) return undefined;
     const documentValidationErrors = propertiesWithErrors.map((property) => {
       const action = actions.find(
-        (action) => property.shape === action.shaclRule,
+        (action) => property?.shape === action.shaclRule,
       );
       return {
         ...property,
--- a/addon/components/document-validation-plugin/card.gts
+++ b/addon/components/document-validation-plugin/card.gts
@@ -37,10 +37,18 @@ export default class DocumentValidationPluginCard extends Component<Sig> {
   }
   get documentValidationErrors() {
     if (!this.validationState) return [];
-    const { propertiesWithErrors } = this.validationState;
+    const { propertiesWithErrors, actions } = this.validationState;
     if (!propertiesWithErrors) return undefined;
-
-    return propertiesWithErrors;
+    const documentValidationErrors = propertiesWithErrors.map((property) => {
+      const action = actions.find(
+        (action) => property.shape === action.shaclRule,
+      );
+      return {
+        ...property,
+        action,
+      };
+    });
+    return documentValidationErrors;
   }
   get propertiesWithoutErrors() {
     if (!this.validationState) return [];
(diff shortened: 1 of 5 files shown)


abeforgit:
About the description: nothing says we can't be _more_ restrictive than the datamodel, right? Idk if there's an extension mechanism in shacl, but otherwise we could just maintain our version of it

addon/components/document-validation-plugin/card.gts
```diff
@@ -114,6 +122,15 @@ export default class DocumentValidationPluginCard extends Component<Sig> {
                 title={{error.subject}}
                 {{on 'click' (fn this.goToSubject error.subject)}}
               >{{t 'document-validation-plugin.see-related-node'}}</AuButton>
+              {{#if error.action}}
+                <AuButton
```
abeforgit, line 126: This work only after clicking twice :/

tests/dummy/app/controllers/besluit-sample.ts
```diff
@@ -355,7 +360,177 @@ export default class BesluitSampleController extends Controller {
...
+		 sh:or (
+      [
+        sh:datatype <http://www.w3.org/2001/XMLSchema#string>;
+      ]
+      [
+        sh:datatype <http://www.w3.org/1999/02/22-rdf-syntax-ns#langString>;
+      ]
+    );
+    sh:minCount 1 ;
+		sh:maxCount 1 ;
```
abeforgit, line 426: formatting - these look like they should be on the same indentation, tripped me up for a second

tests/dummy/app/controllers/besluit-sample.ts
```diff
@@ -355,7 +360,177 @@ export default class BesluitSampleController extends Controller {
...
+        sh:resultMessage "Het besluit mag maximaal één beschrijving hebben.";
```
abeforgit, line 427:
```suggestion
        sh:resultMessage "Het besluit moet exact één beschrijving hebben.";
```

abeforgit: "max 1" is not accurate, it's "exactly 1"
this is otherwise a very weird message when you have a missing description

tests/dummy/app/controllers/besluit-sample.ts
```diff
@@ -355,7 +360,177 @@ export default class BesluitSampleController extends Controller {
...
+              'https://data.vlaanderen.be/shacl/besluit-publicatie#besluit-title-validation',
+            action: (controller: SayController) =>
+              insertTitleAtCursor(controller, this.intl),
+            buttonTitle: 'Insert title',
+          },
+          {
+            shaclRule:
+              'https://data.vlaanderen.be/shacl/besluit-publicatie#besluit-description-validation',
+            action: (controller: SayController) =>
+              insertDescriptionAtCursor(controller, this.intl),
```
abeforgit, line 375: soooo. yeah this simply doesn't work...
The problem is that the rule is checking for exactly 1 description, but the "fix" always inserts a new one. So if you have 2 (easy to do by just copy-pasting, and changing the content of 1 of them so it generates 2 different triples), it still suggests inserting one more...

The setup is simply not granular enough I think... 
Unless there's an easy fix we might have to shelve this ticket and look a this again later

abeforgit (changes requested):

piemonkey (approved):
> @piemonkey I don't know why I can't comment on your question about the descriptions, is a funny one, basically as you have the same text in both of your description nodes there's no problem with it, because they get deduped in the triplestore and you end up having only one description but as soon as you change one you get the error :) I don't really know what to do with that, it can be confusing for the user but the document is technically correct

Hmm, makes sense. I guess the chances of this being a problem in real life are pretty minimal anyway.

All looks good now.

PR #598, opened 2025-09-10
Only auto-focus variables if they contain only a placeholder

piemonkey:
### Overview
The auto-focus behaviour causes problems when a variable has nested nodeviews, e.g. an inline_rdfa inside a variable. This seems to occur in roadsign regulations.

Instead of always focusing the inner editor, only do so for placeholders. All of the variables seem to function correctly after this change.

Gent raising the issue of editing these values: https://binnenland.atlassian.net/browse/GN-5718

### Setup
N/A

### How to test/reproduce
- Insert a variable which contains free text, such as a 'location' (not an oslo_location).
- Click somewhere outside the variable and insert some variables to test.
- Copy-paste the test variables into the location variable.
- Check that editing the text and the variables work as expected.

### Challenges/uncertainties
It's hard to know if there's some other edge-case that is broken by this...

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/clever-beers-rule.md
+++ b/.changeset/clever-beers-rule.md
@@ -1,0 +1,5 @@
+---
+'@lblod/ember-rdfa-editor-lblod-plugins': patch
+---
+
+Do not auto-focus variables unless they only contain a placeholder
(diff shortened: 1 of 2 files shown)


abeforgit:
@piemonkey nice find to just specifically check for that element, I'd have never thought of that

PR #636, opened 2026-04-01
Use pointers to make linking to codelist value URIs possible

piemonkey:
### Overview
See the editor PR for the logic. This is just modifying the codelist logic (and the marcode variable which is just a special case of this).

Jira ticket: https://binnenland.atlassian.net/browse/GN-6037
Editor PR: https://github.com/lblod/ember-rdfa-editor/pull/1367

### Setup
Should be ready to go as it uses a dev build of the editor.

### How to test/reproduce
- Insert a marcode variable, without specifying a value. This node is a plain rdfa node.
- Create a relationship with the marcode node as the 'object'.
- Reloading should maintain these relationships.
- Choose a value for the marcode variable. In the serialised document, there should now be an RDFa triple with the URI of the marcode value as the object.

### Challenges/uncertainties
N/A

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations


abeforgit (changes requested):
This seems to introduce a classic bug: properties that point to literal nodes get doubled on consecutive reloads

reproduction:
- in the test app, "besluit sample" page (or any other page)
- insert decisionTemplate
- check rdfa property viewer in sidebar
- reload the page a few times
- notice properties that point to literalnodes (such as the title and description) get an extra entry each reload

the extra links do point to the same nodes, so the data is likely correct, just something not quite matching up between serialize and parse I think

piemonkey:
> This seems to introduce a classic bug: properties that point to literal nodes get doubled on consecutive reloads

This should be fixed now. I also updated the decision template sample as the title wasn't linked as well as seemingly being quite old.

abeforgit (approved):

PR #637, opened 2026-04-06
Place description context actions

kobemertens:
### Overview
This PR changes the way a location variable is inserted. It will now be inserted as a block_rdfa node instead of an inline variable element. This aligns better with how users think of place descriptions, namely using multiple pre-defined phrases together to form a single place description. To insert these phrases the user can now use the new context actions added in https://github.com/lblod/ember-rdfa-editor/pull/1366 .

JIRA: https://binnenland.atlassian.net/browse/GN-6016
Core contextual actions: https://github.com/lblod/ember-rdfa-editor/pull/1366

### Setup
Link the editor version from (https://binnenland.atlassian.net/browse/GN-6016) to the package.json . Deploy as normal.

### How to test/reproduce
The contextual options should only be available inside a place description block (= location variable). So you will need to insert a mobility measure first. (Or use the variables in the template builder (RB)).
Checklist:
- [ ] RDFa is still correct (`mobiliteit:plaatsbepaling` predicate is still attached).
- [ ] Things still work as expected/desired in the template builder (RB).

### Challenges/uncertainties
Currently, the selected phrase is inserted at the cursor, I feel this is the most predictable/controllable behavior and therefore preferred, however this deviates from the ticket. Feel free to comment on this.

### Checks PR readiness
- [x] changelog
- [x] npm lint


addon/plugins/variable-plugin/actions/create-classic-location-variable.ts
```diff
@@ -88,7 +81,7 @@ export function createClassicLocationVariableAttrs({
     __rdfaId,
     externalTriples,
     backlinks: [...backlinks, ...addedBacklinks],
-    label,
+    label: label ?? 'Plaatsbeschrijving',
```
piemonkey, line 84: I'm always seeing these with a label of locatie. I'm not sure if that's what we want or where that's coming from.

addon/plugins/variable-plugin/contextual-actions/index.ts
```diff
@@ -0,0 +1,184 @@
...
+    const source = node.attrs.source;
+    if (source) {
+      return source;
+    }
+  }
+
+  return fallback;
+}
+
+function getIsZonal(state: EditorState) {
```
piemonkey, line 59: I think we can pull this and the logic in `location/edit.ts isZonal()` into a separate function.

addon/plugins/variable-plugin/contextual-actions/index.ts
```diff
@@ -0,0 +1,184 @@
...
+};
+
+function humanReadableLabel(label: string) {
+  return label.replace(/\$\{[^}]+\}/g, '…');
+}
+
+function createInsertPlaceDescriptionCommand(label: string) {
+  return (state: EditorState, dispatch: (tr: Transaction) => void) => {
+    if (dispatch) {
+      let htmlToInsert = unwrap(label);
```
piemonkey: I think the only thing `unwrap()` does here is crash if `label` is undefined instead of later on in the `wrapVariableInHighlight` fuction.

addon/plugins/variable-plugin/actions/create-classic-location-variable.ts
abeforgit: There was some discussion with the PMs that "Plaatsbeschrijving" would be a better user-facing term

addon/plugins/variable-plugin/contextual-actions/index.ts
abeforgit: yeah according to the types label should never be undefined here so the unwrap should be a noop
if it can be undefined, the arg type should be updated and the function should decide what to do with that case

kobemertens: Ah alright, this was some code I copied from the previous placeholder functionality. Will remove the unwrap.

addon/plugins/variable-plugin/contextual-actions/index.ts
kobemertens: Any suggestions where to put it? I feel like roadsign-regulation-plugin would make sense but I'm not sure if plugins should import from each other?

abeforgit (changes requested):
@kobemertens the contextual actions don't seem to load anymore, not sure what's going on


EDIT: so I did find what's going on, and it's the fact that the button handler now sends a transaction with a `SLASH_COMMANDS_PLUGIN` meta property. 
This implies that the plus button also needs the slash commands plugin to be active, which can be debatable, but in any case, even in #643 (Which I've now merged into this branch) this doesn't work.

This is because the slash-commands plugin does a number of position checks (isInEmptyParagraph, isAtStartOfParagraph, etc) which make sense for typing a slash, but don't necessarily make sense for the button press. 

So some of these return false, and the menu doesn't open. 

For a button press, the menu should **always** open, even if there are no actions available (though it's better to just not show the button in that case of course, which you already do), otherwise it looks broken.

I looks to me like the `slash-commands` plugin is trying to be more like a generic `contextual-actions-menu-state` plugin, so a rename would maybe be a good idea. and then you could distinguish button presses from slash-commands with the right transaction meta values

kobemertens:
> I looks to me like the slash-commands plugin is trying to be more like a generic contextual-actions-menu-state plugin, so a rename would maybe be a good idea. and then you could distinguish button presses from slash-commands with the right transaction meta values.

Yes but the slash-commands plugin also specifically checks for a slash being present, and displays the placeholder message where needed. So a rename would not suffice I think. Maybe we need to split this in 2 plugins, where one just manages the state (listens for 2 transactions `close_context_actions_menu` `open_context_actions_menu`). And have a separate slash-commands plugin, which dispatches transactions to the state-plugin.

kobemertens:
But thinking about this again, I don't think it is the right approach and adds unnecessary complexity. The slash-commands plugin just determines if the menu should be open based on if there is a slash typed or not. We can derive if the menu should be open based on multiple values, and have a separate value for "openness" triggered by the UI button. It should not require a completely new plugin just to manage the menu openness state.

abeforgit:
made a dev release for the fixed editor: `13.7.0-dev.e844abc67c65a776170ec875aa4449277f4de0d6`

addon/plugins/location-plugin/contextual-actions/index.ts
```diff
@@ -0,0 +1,114 @@
...
+      dispatch(tr);
+    }
+    return true;
+  };
+}
+
+export function getContextualActions() {
+  return async function (state: EditorState) {
+    const t = getTranslationFunction(state);
+    // TODO: translate!
```
abeforgit: todo comment

addon/plugins/roadsign-regulation-plugin/actions/insert-measure.ts
```diff
@@ -268,14 +268,16 @@ export default function insertMeasure({
...
 }
 
-function constructMeasureBody(
+function constructMeasureFragment(
   templateString: string,
   variables: Record<string, VariableInstance>,
   schema: Schema,
   backlinks?: IncomingTriple[],
 ) {
+  // TODO: extract this functionality of parsing a text with "variable" placeholders into a fragment to a more general place
```
abeforgit: what's blocking this todo?
I agree it makes sense, and just a function under the roadsign-regulation-plugin folder would be general enough I think

translations/nl-BE.yaml
```diff
@@ -232,7 +232,7 @@ variable-plugin:
   enter-variable-value: Selecteer waarde
   unknown-codelist: Er werden geen mogelijke waarden voor deze codelijst gevonden
   label: Label
-  labelPlaceholder: Type iets...
+  labelPlaceholder: Typ iets…
```
abeforgit: Thanks for cathing the typo, but looking at this now, something like "Vul in..." or "Label invullen..." sounds better 

While gramatically correct, there's something off about "Typ iets" 😬

package.json
```diff
@@ -111,8 +111,7 @@
     "@glint/ember-tsc": "^1.5.0",
     "@glint/environment-ember-loose": "^1.5.0",
     "@glint/environment-ember-template-imports": "^1.5.0",
-    "@glint/template": "^1.7.7",
-    "@lblod/ember-rdfa-editor": "13.6.0",
+    "@lblod/ember-rdfa-editor": "13.7.0",
```
abeforgit: I think this needs 13.7.1 now that its out, but more importantly: you need to update the peerDeps to require that version

Which would make this a breaking release, which is OK for this repo (cause its a bit of a stupid setup, ideally each plugin should be its own package that can evolve its version independently, but yeah)

abeforgit (changes requested):

addon/plugins/variable-plugin/contextual-actions/index.ts
abeforgit: they can import from eachother, it's a monolith package anyway (but even if it wouldnt be, that'd still be ok. It's just javascript)

translations/nl-BE.yaml
kobemertens: [390f25a](https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/637/commits/390f25ac081ab74155bccbb5cfeb7b4d51c0eaf6)

addon/plugins/roadsign-regulation-plugin/actions/insert-measure.ts
kobemertens: [359a77d](https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/637/commits/359a77dbdd70609a52aed9e55c63a2cef497efea)

package.json
kobemertens: [d2d939b](https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/637/commits/d2d939bd5fe39e51146ad1df777aa48d6f54236b)

abeforgit:
@kobemertens nice job on this!

PR #640, opened 2026-04-14
Add optional explicit subject to link locations to

abeforgit:
### Overview

For usecases like cipal, there is no relevant subject present to link the location to, since their documents do not contain the decision or article nodes.
So this pr adds a config option which allows you to link all locations to an explicit subject, in a similar way as the decision-topic plugin and others work.

As part of the process, I refactored the replaceSelectionWithLocation function to take named args. That's a separate commit, so you can see meaningful diffs if you filter it out.

https://binnenland.atlassian.net/browse/GN-6074

https://github.com/lblod/frontend-embeddable-notule-editor/pull/288

### Setup

`pnpm start`

### How to test/reproduce

The besluit-sample page of the dummy app has an explicit subject configured as `http://example.org/foo-decision`

Check that all operations with locations produce a triple which links the location to that subject using the prov:atLocation predicate.

It should also not intervene with the subjectTypesToLinkTo logic
### Checks PR readiness

- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/addon/components/location-plugin/insert.gts
+++ b/addon/components/location-plugin/insert.gts
@@ -283,13 +283,13 @@ export default class LocationPluginInsertComponent extends Component<Signature> 
         // if they exist
         const newExternalTriples: FullTriple[] = (
           locNode.attrs.externalTriples as FullTriple[]
-        ).slice();
-        for (const tr of newExternalTriples) {
+        ).map((tr) => {
           if (PROV('atLocation').matches(tr.predicate)) {
-            tr.object = df.namedNode(toInsert.uri);
+            return { ...tr, object: df.namedNode(toInsert.uri) };
+          } else {
+            return tr;
           }
-        }
-        // update the location value and the external triples
+        }); // update the location value and the external triples
         this.controller.withTransaction((tr) => {
           return transactionCombinator(
             this.controller.activeEditorState,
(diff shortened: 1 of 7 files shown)


addon/components/location-plugin/insert.gts
```diff
@@ -274,13 +277,27 @@ export default class LocationPluginInsertComponent extends Component<Signature>
...
+        const df = new SayDataFactory();
+        const { pos, value: locNode } = this.selectedLocationNode;
+        // The location's subject has likely changed, so we should update the external links too
+        // if they exist
+        const newExternalTriples: FullTriple[] = (
+          locNode.attrs.externalTriples as FullTriple[]
+        ).slice();
+        for (const tr of newExternalTriples) {
+          if (PROV('atLocation').matches(tr.predicate)) {
+            tr.object = df.namedNode(toInsert.uri);
```
piemonkey: Turns out my distaste at seeing mutations isn't so irrational after all... It looks like since we're modifying the objects in the array, and so modifying the contents of the original array, this means the new value is actually kept after an undo.
Repro: 
- Insert a location, store the URI somewhere.
- Update the location, store the new URI somewhere (you might need to click away and select the location again).
- Hit undo, look at the rdfatools and see that the subject has been reverted but the external triple still references the second location.

piemonkey (changes requested):
Looks good in general, just one very obscure bug.

addon/components/location-plugin/insert.gts
abeforgit: huh wut, but I explicitly made a copy with slice for that reason :thinking: I'll look into it

abeforgit: ooh its not a deep copy... :man_facepalming:

abeforgit: great catch

piemonkey (approved):

PR #644, opened 2026-04-21
feature/glint V2

kobemertens:
### Overview
Upgrade to glint V2

### Checks PR readiness
- [x] changelog
- [x] npm lint


piemonkey (commented):
Running typechecks from the command line work fine, but I see errors in my editor when looking at gts files that I don't see in other glint v2 repos.

Decorators (such as `@action` and `@tracked` at least) are shown having the error:
`Unable to resolve signature of method decorator when called as an expression.`

Imports from `@lblod/ember-rdfa-editor-lblod-plugins` say they cannot find the module.

Maybe this is a problem with my neovim setup, but since e.g. the editor repo works, that seems strange.

tsconfig.json
```diff
@@ -2,10 +2,8 @@
 {
   "extends": "./tsconfig.main.json",
   "compilerOptions": {
-    "rootDir": "."
+    "rootDir": ".",
+    "lib": ["dom", "dom.iterable", "es2022"]
```
piemonkey, line 6: It seems strange that we need this here but not in the editor repo. I guess because the editor tsconfig extends `@tsconfig/ember/tsconfig.json`, but this does not? Maybe it would be best to follow that more standard (at least as far as I know) approach?

piemonkey: Hmm, this actually surfaces a load more possible errors in the types, so for that we'd have to explicitly turn off those rules and then ideally fix them and enable them later...

piemonkey:
Sorry for the chaotic comments...
I just realised that while running `pnpm start`, I get a build error as well as seeing a bunch of typechecking errors about failing to find modules. The error is
```
Build Error (broccoli-persistent-filter:Babel > [Babel: ember-power-select]) in ember-power-select/components/power-select-multiple/input.ts

[BABEL]: Cannot find module '@babel/helper-module-imports'
```

Update: Doing a `git clean` clears the error (how`‽). I'm still seeing errors when running `pnpm typecheck` (but not `pnpm lint:types`).

tsconfig.json
kobemertens: We do this in the template-export-service as well, but extending from `@tsconfig/ember/tsconfig.json` might be better indeed

kobemertens:
Superseded by https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/647

PR #647, opened 2026-05-04
add hybrid glintv2 setup

kobemertens:
Glint V2 does not support handlebars, so we need a hybrid setup. I follow the steps in the migration guide to this point: https://github.com/typed-ember/glint/blob/main/docs/v2-upgrade.md#milestone-hybrid-v1--v2-glint-installation

### Overview
Moves to the hybrid setup described in https://github.com/typed-ember/glint/blob/main/docs/v2-upgrade.md#milestone-hybrid-v1--v2-glint-installation


piemonkey (approved):
As discussed, glintV2 seems to be preferable to handlebars types, so seems good to me.

PR #651, opened 2026-05-12
Add function to get oslo locations from document

lagartoverde:
### Overview
Added a function that scans the document for all the oslo locations and returns the place/address/area associated with them

https://binnenland.atlassian.net/browse/GN-6104

### Setup
None

### How to test/reproduce
To be tested with Kobe work but you can add it to any button and check that it returns the appropiate data

### Challenges/uncertainties
None

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/addon/plugins/location-plugin/utils/get-document-locations.ts
+++ b/addon/plugins/location-plugin/utils/get-document-locations.ts
@@ -1,15 +1,68 @@
 import { SayController } from '@lblod/ember-rdfa-editor';
+import { Area, Place } from './geo-helpers';
+import { Address } from './address-helpers';
+
+type locationsWithDistanceType = {
+  location: Place | Address | Area;
+  distance: number;
+};
+
+type locationMetadataType = {
+  [locationUri: string]: { ocurrences: number; distance: number };
+};
 
 export default function getDocumentLocations(controller: SayController) {
   const state = controller.mainEditorState;
   const doc = state.doc;
-  const locations: (Place | Address | Area)[] = [];
-  doc.descendants((node) => {
+  const locationsWithDistance: locationsWithDistanceType[] = [];
+  const { selection } = controller.mainEditorState;
+  doc.descendants((node, pos) => {
     if (node.type.name === 'oslo_location') {
-      locations.push(node.attrs.value);
+      const distance = Math.abs(pos - selection.from);
+      locationsWithDistance.push({
+        location: node.attrs.value,
+        distance: distance,
+      });
       return false;
     }
     return true;
   });
+  const locationMetadata: locationMetadataType = {};
+  const locationsDedup: locationsWithDistanceType[] = [];
+  for (const locationWithDistance of locationsWithDistance) {
+    const uri =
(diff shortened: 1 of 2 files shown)


abeforgit:
@lagartoverde regardless of whether we merge this into master, once kobe confirms the function does what it needs to do, you can put the jira ticket to done (don't forget your burned SPs)

@kobemertens would appreciate you reviewing this asap so we can still clock it as part of this sprint

addon/plugins/location-plugin/utils/get-document-locations.ts
```diff
@@ -0,0 +1,64 @@
+import { SayController } from '@lblod/ember-rdfa-editor';
+import { Area, Place } from './geo-helpers';
+import { Address } from './address-helpers';
+
+type locationsWithDistanceType = {
```
kobemertens: I think this needs to start with a capital letter

addon/plugins/location-plugin/utils/get-document-locations.ts
```diff
@@ -0,0 +1,64 @@
...
+  location: Place | Address | Area;
+  distance: number;
+};
+
+type locationMetadataType = {
```
kobemertens: This one as well

addon/plugins/location-plugin/utils/get-document-locations.ts
```diff
@@ -0,0 +1,64 @@
...
+  [locationUri: string]: { ocurrences: number; distance: number };
+};
+
+export default function getDocumentLocations(controller: SayController) {
```
kobemertens: The function should take a `EditorState` instead according to the current API we have

kobemertens (changes requested):
Looks good, just some minor remarks.

addon/plugins/location-plugin/utils/get-document-locations.ts
```diff
@@ -0,0 +1,63 @@
...
+    ) {
+      return locationMetadata[uriA].distance - locationMetadata[uriB].distance;
+    } else {
+      return (
+        locationMetadata[uriB].ocurrences - locationMetadata[uriA].ocurrences
+      );
+    }
+  });
+  return locations;
+}
```
kobemertens, line 63: I feel like some kind of scoring based on occurences and distance might be better here instead of sort on occurences first and then on distance, not sure if you'd be down to look into that? Otherwise we'll look at it if we have more concrete examples of documents where this will be used.

kobemertens (approved):

addon/plugins/location-plugin/utils/get-document-locations.ts
lagartoverde: That would be interesting but I don't think we should paralize this ticket for it, we can go for this, and latter modify it when the heuristic is more clear.
Do you want to merge this in your branch better than master? this way you can implement and merge as part of your ticket

kobemertens: If it's up to me, I would merge to master, then I can merge my branch with master.

PR #654, opened 2026-05-19
Location context actions

kobemertens:
### Overview
When highlighting a location node, we want to be able to use the context menu to quickly change/insert a location. This PR adds buttons to the context actions to open the modal in either place, address or area mode.

https://binnenland.atlassian.net/browse/GN-6003

### Setup
Depends on this PR for the editor: https://github.com/lblod/ember-rdfa-editor/pull/1374

### How to test/reproduce
In the decision-sample, clicking a location node (or location placeholder), should open the context actions with 3 options: Insert address, insert point on the map, insert area. 

### Challenges/uncertainties
Currently the context actions get shown each time a location node is selected, this can happen not only trough clicking on the node but also undo-ing, closing the modal, ... This might not be desired. It is not possible anymore to change the content of the Node anymore because clicking it opens the context actions and moves the cursor to the search field.

Also, when the context actions are closed, to reopen a user cannot click the Location Node again because the reactivity system triggers on selection change. I think it is important that this is fixed, although maybe not in this PR, but rather when we look at the rework of this component in the next sprint.

### Checks PR readiness
- [x] changelog
- [x] npm lint

### DIFF
--- a/addon/components/location-plugin/insert.gts
+++ b/addon/components/location-plugin/insert.gts
@@ -47,7 +47,7 @@ import {
 export type CurrentLocation = Address | GlobalCoordinates | undefined;
 
 function updateFromNode<T extends Address | Place | Area, R>(
-  Class: new (...args: unknown[]) => T,
+  Class: new (...args: never[]) => T,
   extractFunc: (current: T) => R,
   defaultValue?: R,
 ) {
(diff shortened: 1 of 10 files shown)


kobemertens:
Waiting for the AUv4 version of this repo. https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/657

addon/components/location-plugin/insert.gts
```diff
@@ -38,11 +38,16 @@ import type { Option } from '@lblod/ember-rdfa-editor-lblod-plugins/utils/option
...
+  closeLocationModal,
+  getLocationModalsPluginState,
+  openLocationModal,
+} from '@lblod/ember-rdfa-editor-lblod-plugins/plugins/location-plugin';
 
 export type CurrentLocation = Address | GlobalCoordinates | undefined;
 
 function updateFromNode<T extends Address | Place | Area, R>(
-  Class: new (...args: unknown[]) => T,
+  Class: new (...args: never[]) => T,
```
abeforgit, line 50: this is a weird change, what's the reason?

PR #656, opened 2026-05-27
Feature/locked placeholder

lagartoverde:
### Overview
Added the new locked placeholder feature, this consists of 2 new nodes, a card to insert locked placeholders and a function to replace them, that works with the processDocumentHeadlessly function

https://binnenland.atlassian.net/browse/GN-6126

### Setup
You can replace any button in the editor to run your replace function in order to test it,
```
const values = {
      block: '<ul><li>Hello</li></ul>',
      inline: '<b>Inline content</b>',
    };
    this.controller.doCommand((state, dispatch) => {
      if (!dispatch) {
        return false;
      }
      const { result, transaction } = replaceLockedPlaceholderContent(
        state,
        values,
      );
      if (result.every((ok) => ok)) {
        dispatch(transaction);
        return true;
      }
      return false;
    });
 ```
 This is what I was doing while testing, you can use the keys and values that you want and you can try constructing prosemirror nodes by passing down a function that takes the state as an argument

### How to test/reproduce
Verify that you can add both types of placeholders, via the card and by editing the html directly and that you can replace them

### Challenges/uncertainties
None

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/addon/components/locked-placeholder-plugin/insert.gts
+++ b/addon/components/locked-placeholder-plugin/insert.gts
@@ -1,0 +1,140 @@
+import Component from '@glimmer/component';
+import { on } from '@ember/modifier';
+import AuButton from '@appuniversum/ember-appuniversum/components/au-button';
+import AuCard from '@appuniversum/ember-appuniversum/components/au-card';
+import AuHeading from '@appuniversum/ember-appuniversum/components/au-heading';
+import AuContent from '@appuniversum/ember-appuniversum/components/au-content';
+import AuFormRow from '@appuniversum/ember-appuniversum/components/au-form-row';
+import AuLabel from '@appuniversum/ember-appuniversum/components/au-label';
+import AuNativeInput from '@lblod/ember-rdfa-editor-lblod-plugins/components/au-native-input';
+import AuRadioGroup from '@appuniversum/ember-appuniversum/components/au-radio-group';
+import { NodeSelection, SayController } from '@lblod/ember-rdfa-editor';
+import { type LocationPluginConfig } from '@lblod/ember-rdfa-editor-lblod-plugins/plugins/location-plugin/node';
+import { action } from '@ember/object';
+import t from 'ember-intl/helpers/t';
+import { tracked } from '@glimmer/tracking';
+
+interface Signature {
+  Args: {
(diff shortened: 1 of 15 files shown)

### CODE REVIEW

File: addon/components/locked-placeholder-plugin/insert.gts [hunk 1]
```diff
@@ -0,0 +1,161 @@
...
+    this.label = (event.target as HTMLInputElement).value;
+  }
+
+  @action
+  updateType(type: string) {
+    this.type = type;
+  }
+
+  <template>
+    {{! @glint-nocheck: not typesafe yet }}
```
abeforgit: this should not be here, this template is typesafe
If you're getting errors due to appuniversum, solve or ignore them individually

addon/plugins/locked-placeholder-plugin/nodes/block-locked-placeholder.ts
```diff
@@ -0,0 +1,68 @@
...
+const emberNodeConfig: EmberNodeConfig = {
+  name: 'block-locked-placeholder',
+  component: BlockLockedPlaceholderNodeView as unknown as ComponentLike,
+  inline: false,
+  group: 'block',
+  content: 'inline*',
+  atom: true,
+  draggable: false,
+  needsFFKludge: true,
+  editable: true,
```
abeforgit, line 52: I think the "atom" property overrides this, but still its better to set this to false, as that's the intention of this node

addon/plugins/locked-placeholder-plugin/nodes/inline-locked-placeholder.ts
```diff
@@ -0,0 +1,68 @@
...
+    },
+    0,
+  ];
+};
+
+const emberNodeConfig: EmberNodeConfig = {
+  name: 'inline-locked-placeholder',
+  component: InlineLockedPlaceholderNodeView as unknown as ComponentLike,
+  inline: true,
+  group: 'inline variable',
```
abeforgit: why is this in the variable group?

abeforgit (changes requested):
small nitpicks I noticed, still have to test

PR #662, opened 2026-06-11
always use text content to avoid parsing problems

lagartoverde:
### Overview
Seems like title parsing on the structures is failing, due to including the whole span in the prosemirror json. I added the text content accessor so it always pass down a string, this works on all the structures I tested, but if you make it break please let me know as this is a risky PR

https://binnenland.atlassian.net/browse/GN-6161

### Setup
None

### How to test/reproduce
Check that all the structures are parsed correctly even if they have titles.
If you want to really make it break before the PR link to GN and open the RPR - rechtspositieregeling 2026 template and try to save

### Challenges/uncertainties
None

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/addon/plugins/structure-plugin/node.ts
+++ b/addon/plugins/structure-plugin/node.ts
@@ -211,7 +211,7 @@ export const emberNodeConfig: (
       const hasTitle = node.attrs.hasTitle as boolean;
       const titleHTML = hasTitle
         ? parser.parseFromString(node.attrs.title, 'text/html').body
-            .firstElementChild
+            .firstElementChild?.textContent
         : null;
       const headerFormat = node.attrs.headerFormat as string;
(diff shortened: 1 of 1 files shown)


abeforgit (changes requested):
hmm
The reason we're parsing the element here is because titles can have formatted text, such as bold or underlined words, or even highlighted sections. 

With this we would lose that capability. That would maybe be acceptable, but I don't see how this can be the real cause of the issue. If you use structure titles on master in the dummy test app, and reload them, they work fine, even if they have fancy formatting. So I'd like to know how exactly this is failing for that particular template before we commit to a solution like that

abeforgit:
@lagartoverde I think we can close this without merging right?

lagartoverde:
> @lagartoverde I think we can close this without merging right?

yep

PR #674, opened 2026-08-05
Update snippet list names when updated on template builder

piemonkey:
### Overview
Load snippet list names from a sparql query instead of storing them on the node attrs. Adds a service that caches these results and also use it to cache snippets.

Part of https://binnenland.atlassian.net/browse/GN-6182

### Setup
Needs to be pointing to an RB where you can change snippet list names (e.g. dev)

### How to test/reproduce
- In the regulatory statements test page, insert snippet placeholders.
- Refresh to ensure the doc has been put into local storage, then switch to the besluit page. (this split is due to an old confusion between 'regulatory statements' the part of GN and 'regulatory statements' the template builder app, so in this case, this represents making a template and consuming one, whereas most of the config reflects the different part of GN)
- Change the name of one of the snippet lists, then reload the plugins. Note that the list name is updated in the editor.
- Opening the 'insert snippet' modal for a particular set of lists should only result in sparql queries per page, and should be cached after that.
- All of the usual snippet functionality works as expected.

### Challenges/uncertainties
N/A

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations


abeforgit:
works well, but I have my doubts on the failure case. If the name cannot be loaded, this is what it shows (you can test this by blocking the domain in the devtools)
image


Which isn't bad, but perhaps a failure indication would be better 

This is also what it's rendering while loading, but that's usually so fast it's not much more than a flicker. Not sure if a spinner would be better or worse for these very fast loads, but a I guess it can always occur that the load is slow. I'll merge this and add a ticket for these enhancements to avoid blocking the story.

